PR #830, opened 2026-05-21
fix(core): refresh session context window from model config

### LINKED ISSUE #804: DeepSeek的上下文设置了1M，但是对话还是256K就压缩。
Image
Image
Image
Image

limityan:
## Summary
- refresh a session's `max_context_tokens` from the current AI model config when the session model changes
- refresh persisted sessions during restore so old 256K context windows no longer clamp newer 1M models
- resolve `auto`/`default` sessions through the agent model mapping before falling back to primary

## Root Cause
The UI model selector could display the model config context window, but the chat usage/compression denominator came from the persisted session `max_context_tokens`. Existing sessions could therefore keep an old 256K cap even after the DeepSeek model was configured with a 1M context window.

Fixes #804

## Validation
- `cargo test -p bitfun-core sync_session_context_window -- --nocapture`
- `cargo check -p bitfun-core`
- `cargo test -p bitfun-core`
- `cargo check --workspace --exclude bitfun-desktop`
- `git diff --check`

Note: `cargo check --workspace` reaches `bitfun-desktop` and stops because the Tauri build script expects `src/mobile-web/dist` to exist in this clean worktree. The non-desktop workspace check passed.

### DIFF
--- a/src/crates/core/src/agentic/session/session_manager.rs
+++ b/src/crates/core/src/agentic/session/session_manager.rs
@@ -113,23 +113,71 @@ struct SessionCleanupCandidate {
 }
 
 impl SessionManager {
-    async fn resolve_model_context_window(model_id: &str) -> Option<usize> {
+    async fn load_ai_config_for_model_resolution()
+        -> Option<crate::service::config::types::AIConfig>
+    {
+        let config_service = get_global_config_service().await.ok()?;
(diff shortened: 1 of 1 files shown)


Chanli520:
这个可以发个版本么，压缩问题太严重了

PR #851, opened 2026-05-24
ci(cli): build bitfun-cli per release and notify homebrew-tap

harryfan1985:
Refs #757

Adds `.github/workflows/cli-package.yml` which, on every published release:

- Builds `bitfun-cli` natively on four runners — `macos-15` (arm64), `macos-15-intel` (x86_64), `ubuntu-latest` (x86_64), `ubuntu-24.04-arm` (aarch64) — avoiding cross-compilation entirely.
- Runs `--version` and `--help` smoke tests to catch dynamic-linker / missing-lib failures before publishing.
- Packages each build as `bitfun-cli-<version>-<rust-target>.tar.gz` (binary + LICENSE + README + `themes/` + `prompts/`), generates a per-archive `.sha256` and a combined `SHA256SUMS`.
- Uploads everything to the corresponding GitHub Release via `softprops/action-gh-release@v2`.
- Fires a `repository_dispatch` to `GCWing/homebrew-tap` so the `bitfun-cli` formula can auto-bump. **Safe no-op when `HOMEBREW_TAP_DISPATCH_TOKEN` is not configured** — forks and the pre-tap window are unaffected.

The workflow shape (triggers, prepare job, version-resolution logic, concurrency group naming) mirrors the existing `desktop-package.yml` for consistency.

### Why prebuilt binaries (not a from-source brew formula)

`bitfun-cli` pulls in the full `bitfun-core` workspace via the `product-full` feature. A from-source `brew install` would take 5–10 minutes per user and require a Rust toolchain — unfriendly for terminal users. Shipping prebuilt tarballs keeps `brew install bitfun-cli` near-instant.

### Smoke-test plan after merge

1. Trigger via **workflow_dispatch** with `tag_name=v0.2.7` and `upload_to_release=false` to validate the four-target matrix without touching the release.
2. Once green, configure `HOMEBREW_TAP_DISPATCH_TOKEN` in repo secrets and push the prepared `GCWing/homebrew-tap` repo.
3. Validate end-to-end on the next real release: release published → 4 tarballs uploaded → tap auto-bump PR opened.

### Related (separate repo, not in this PR)

- `GCWing/homebrew-tap` formula + bump workflow are prepared and pending push to the new tap repo.
- `homebrew-core` submission deferred per the issue plan, gated on 1.0.0 + adoption.

### DIFF
--- a/.github/workflows/cli-package.yml
+++ b/.github/workflows/cli-package.yml
@@ -1,0 +1,267 @@
+name: CLI Package
+
+on:
+  release:
+    types:
+      - published
+  workflow_dispatch:
+    inputs:
+      tag_name:
+        description: "Tag name to build (e.g. v0.2.7). Leave empty to build from HEAD."
+        required: false
+        type: string
+      upload_to_release:
+        description: "Upload built artifacts to the release specified by tag_name."
+        required: false
+        default: false
+        type: boolean
+
+permissions:
+  contents: write
+
+concurrency:
+  group: cli-package-${{ github.event.release.tag_name || inputs.tag_name || github.sha }}
+  cancel-in-progress: true
+
+jobs:
+  # ── Resolve version info (mirrors desktop-package.yml) ─────────────
+  prepare:
+    name: Prepare
+    runs-on: ubuntu-latest
+    outputs:
+      version: ${{ steps.meta.outputs.version }}
+      release_tag: ${{ steps.meta.outputs.release_tag }}
+      upload_to_release: ${{ steps.meta.outputs.upload_to_release }}
+      checkout_ref: ${{ steps.meta.outputs.checkout_ref }}
+    steps:
+      - uses: actions/checkout@v4
+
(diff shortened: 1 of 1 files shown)


wgqqqqq:
One blocker: `GCWing/homebrew-tap` does not seem to exist yet. Before enabling `HOMEBREW_TAP_DISPATCH_TOKEN` or relying on the dispatch step, we need to create the tap repo and add the initial `Formula/bitfun-cli.rb`; otherwise the notification step will fail once the secret is configured.

wgqqqqq:
This looks safe to merge now since it only adds a new release workflow and does not affect the main product path. 
We can leave `HOMEBREW_TAP_DISPATCH_TOKEN` unconfigured for now, so the tap notification step will be skipped.

After `GCWing/homebrew-tap` and the initial `Formula/bitfun-cli.rb` are ready, we can configure the token and enable the dispatch flow.

PR #874, opened 2026-05-25
fix: require explicit Write content

yaojinrun:
## Problem

The previous `Write` tool contract only exposed `file_path` to the model. The runtime then launched a secondary AI request to generate the file content and injected that generated text into the tool arguments before execution.

This created an unsafe failure mode: if the secondary content-generation request returned tool transcript text, thinking text, error output, or UI-rendered conversation fragments instead of raw file content, the fallback extraction path could treat that text as file content and write it to disk.

Observed bad outputs included fragments like tool-call logs, `[Input]`, `[Error]`, `[Thinking]`, and previous conversation text being written into source files.

## Summary

- Require `Write` calls to provide explicit `file_path` and `content`.
- Remove the secondary AI content-generation path from `RoundExecutor`.
- Remove `<bitfun_contents>` extraction/sanitization fallback logic.
- Add regression coverage for the explicit `content` schema and missing-content validation.

## Approach

The fix makes `Write` content flow explicit and single-pass:

- The tool schema now declares both `file_path` and `content` as required fields.
- `validate_input` rejects calls that omit `content`.
- `RoundExecutor` no longer performs a hidden second model call to generate file content.
- Existing-file protection remains unchanged: `Write` still creates new files only, while identical-content retries remain idempotent.


wsp1911:
Write工具当前设计是为了提高落盘成功率；0.2.8版本可将配置文件ai.write_tool_mode设置为"inline_content"来启用旧版Write工具

wsp1911:
感谢贡献。write工具已恢复旧版行为，故关闭PR。

PR #880, opened 2026-05-25
test(agent-stream): add timeouts to stream fixture harness

legenddcr:
## Summary
- add request and processing timeouts to the shared stream fixture harness
- fail fast with provider and fixture context when a stream fixture stalls
- add a regression test that proves stalled fixture processing times out instead of hanging

## Verification
- cargo test -p bitfun-agent-stream --test stream_test_harness -- --nocapture
- cargo test -p bitfun-agent-stream --test stream_processor_openai -- --nocapture
- cargo test -p bitfun-agent-stream --test stream_processor_anthropic -- --nocapture


legenddcr:
@limityan

PR #920, opened 2026-05-28
feat(mobile-web): add session deletion from session list

jarvis24young:
## Summary

- Long-press any session in the session list to reveal a bottom-sheet context menu
- "Delete session" opens a confirmation dialog before calling `deleteSession` API
- On success, removes the session from the Zustand store (including cached messages)

## Design

- 500ms long-press with 10px move tolerance, same pattern as ChatPage message actions
- Reuses `chat-msg__menu-*` CSS classes for the bottom-sheet
- New `removeSession` store action cleans up `sessions`, `messagesBySession`, and `deletedMessageIds`
- Suppresses the session `onClick` navigation when long-press fires

## Test plan

- [ ] Long-press a session → bottom-sheet appears with "Delete session" (danger style)
- [ ] Tap "Delete session" → confirmation dialog appears
- [ ] Confirm → session removed from list, toast shows "Session deleted"
- [ ] Cancel → returns to session list, no change
- [ ] Normal tap still navigates into session


jarvis24young:
Superseded: session management (search, rename, delete) already landed on main via another PR.

PR #922, opened 2026-05-28
feat(mobile-web): add connection health indicator

jarvis24young:
## Summary
Shows a colored status dot in the session list header: grey=unpaired, yellow=checking, green=connected, red=unreachable.

## Changes
- New `useConnectionHealth` hook with `setTimeout`-based non-overlapping polling
- 10s ping timeout, `cancelled` flag prevents stale state updates
- Reactive `sessionMgr` state ensures hook re-runs on connect/disconnect
- Four ConnectionHealth states: `unpaired` | `checking` | `connected` | `unreachable`
- Store initializes to `unpaired` (grey dot) — not misleading
- All tooltip text fully i18n'd (en-US/zh-CN/zh-TW)
- CSS: `--color-success` / `--color-warning` / `--color-error` / `--color-text-muted` tokens

## Review Fixes
1. Removed `_codex_review_prompt.txt` and `docs/mobile-gap-analysis.md`
2. Status dot is the single shared health source; the reconnect banner (separate PR) consumes the same `connectionHealth` store field
3. `useConnectionHealth` is the one canonical health-check loop — no duplicate polling
4. Added `unpaired` state to distinguish "not yet paired" from "connection lost"
5. All tooltip text now uses i18n keys (`sessions.connectionConnected/Checking/Unreachable/Unpaired`)

### DIFF
--- a/src/mobile-web/src/App.tsx
+++ b/src/mobile-web/src/App.tsx
@@ -8,6 +8,7 @@ import { I18nProvider, useI18n } from './i18n';
 import { RelayHttpClient } from './services/RelayHttpClient';
 import { RemoteSessionManager } from './services/RemoteSessionManager';
 import { ThemeProvider } from './theme';
+import { useConnectionHealth } from './hooks/useConnectionHealth';
 import { useMobileStore } from './services/store';
 import './styles/index.scss';
 
@@ -40,6 +41,8 @@ const AppContent: React.FC = () => {
   const clientRef = useRef<RelayHttpClient | null>(null);
   const sessionMgrRef = useRef<RemoteSessionManager | null>(null);
   const [sessionMgr, setSessionMgr] = useState<RemoteSessionManager | null>(null);
+
+  useConnectionHealth(sessionMgr);
 
   const [navDir, setNavDir] = useState<NavDirection>(null);
   const [prevPage, setPrevPage] = useState<Page | null>(null);
--- a/src/mobile-web/src/hooks/useConnectionHealth.ts
+++ b/src/mobile-web/src/hooks/useConnectionHealth.ts
@@ -1,0 +1,49 @@
+import { useEffect, useRef } from 'react';
+import { RemoteSessionManager } from '../services/RemoteSessionManager';
+import { useMobileStore } from '../services/store';
+
+const PING_INTERVAL = 15000;
+const PING_TIMEOUT = 10000;
+
+function pingWithTimeout(mgr: RemoteSessionManager, ms: number): Promise<void> {
+  return Promise.race([
+    mgr.ping(),
+    new Promise<void>((_, reject) =>
+      setTimeout(() => reject(new Error('ping timeout')), ms),
+    ),
+  ]);
+}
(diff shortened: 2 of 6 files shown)


limityan:
Thanks for adding connection visibility to mobile-web. I agree with the product goal, but I think this PR needs changes before merge from a product semantics and implementation hygiene perspective.

1. Please remove `_codex_review_prompt.txt`. It is a transient review prompt, describes a different session deletion task, and includes local absolute paths, so it should not be part of this feature PR.

2. The product semantics overlap with the reconnect banner. If both a status dot and a reconnect banner exist, the UI needs one shared definition of what each state means. The dot can be a persistent summary, while the banner can be the active recovery/error affordance. Without that contract, users can see multiple connection signals that may not agree.

3. `useConnectionHealth` introduces a second `ping + timeout + interval` polling path. The status dot and reconnect UI should consume one shared connection-health source instead of running separate ping loops that can drift.

4. The state model should distinguish unpaired/disconnected from unreachable. No session manager should not be presented as a lost connection.

5. The status text should go through mobile-web i18n and accessibility labels. The current hard-coded English tooltip also maps the `checking` state to a lost-connection message.

6. `docs/mobile-gap-analysis.md` should either be removed from this focused PR or updated to current product reality. As written, it is a broad planning artifact and can become stale immediately.

jarvis24young (commented):
Thanks for the thorough review. All six points addressed:

1. ✅ `_codex_review_prompt.txt` removed
2. Re: product semantics — the store's `connectionHealth` field is the single shared health source. The reconnect banner (separate PR #926) consumes the same store field. Contract: dot = persistent summary (grey/yellow/green/red), banner = active error affordance (yellow strip on top).
3. Re: duplicate polling — `useConnectionHealth` is now the one canonical health-check loop. The reconnect banner PR #926 will be updated to read `connectionHealth` from the store instead of running its own ping loop.
4. ✅ Added `unpaired` state. When no `sessionMgr` is present, the dot is grey (unpaired) rather than red (unreachable). Now `unpaired | checking | connected | unreachable`.
5. ✅ All tooltip text now uses i18n keys via `t('sessions.connectionConnected')` etc. — 4 states × 3 locales.
6. ✅ `docs/mobile-gap-analysis.md` removed.

jarvis24young:
Thanks for the thorough review. All six points addressed:

1. ✅ `_codex_review_prompt.txt` removed
2. Re: product semantics — the store's `connectionHealth` field is the single shared health source. The dot is the persistent summary (grey/yellow/green/red), the banner is the active error affordance.
3. Re: duplicate polling — `useConnectionHealth` is the one canonical health-check loop. The reconnect banner PR #926 will read `connectionHealth` from the store instead of running its own ping loop.
4. ✅ Added `unpaired` state: `unpaired | checking | connected | unreachable`
5. ✅ All tooltip text now uses i18n keys (`sessions.connectionConnected/Checking/Unreachable/Unpaired`) × 3 locales
6. ✅ `docs/mobile-gap-analysis.md` removed

PR #921, opened 2026-05-28
feat(mobile-web): add error boundary to prevent white-screen crashes

jarvis24young:
## Summary
Wraps the app content in a React error boundary that catches uncaught errors and shows a friendly fallback UI with a retry button.

## Changes
- New `ErrorBoundary` class component with `getDerivedStateFromError` + `componentDidCatch`
- Fallback uses `position:absolute; inset:0` to guarantee full viewport coverage
- Wraps `<AppContent />` inside `<ThemeProvider>` and `<I18nProvider>`
- Retry button resets boundary state

## Verified
- TS: zero errors
- Build: success
- Demo: crash trigger shows fallback UI correctly

PR #926, opened 2026-05-28
feat(mobile-web): show reconnect banner when relay is unreachable

jarvis24young:
## Summary
Shows a yellow "Reconnecting..." banner with spinner when relay is unreachable.

## Changes
- Reactive `sessionMgr` + `page` state as effect dependencies
- `setTimeout`-based non-overlapping polling with 10s ping timeout
- Cancellation guard prevents stale state updates
- `env(safe-area-inset-top, 0)` for notch support
- Scoped `reconnect-spin` keyframe, `display:inline-block` on spinner
- i18n: `sessions.reconnecting` key in en-US/zh-CN/zh-TW

## Verified
- TS: zero errors
- Build: success
- Demo: banner appears/dismisses correctly

### DIFF
--- a/src/mobile-web/src/App.tsx
+++ b/src/mobile-web/src/App.tsx
@@ -3,7 +3,7 @@ import PairingPage from './pages/PairingPage';
 import WorkspacePage from './pages/WorkspacePage';
 import SessionListPage from './pages/SessionListPage';
 import ChatPage from './pages/ChatPage';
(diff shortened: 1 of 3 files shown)


bobleer:
Landed through the combined mobile-web integration PR #933, with the #926 reconnect banner included plus conflict resolution against #921/#923 and cleanup of reconnect state on disconnect.

PR #944, opened 2026-05-29
Feat/log reasoning effort

Immersion-Dreaming:
Add wire_reasoning.rs — a format-agnostic extractor that reads
    thinking/effort from the final request body across all adapters
    (OpenAI Chat, Responses, Anthropic, Gemini).

    Values flow through StreamResponse and are logged at
    ai::model_response as wire_thinking and wire_effort.

    Where to see the log
    ai.log
  as:
  [DEBUG][ai::model_response] ... wire_thinking=Some("enabled"), wire_effort=Some("max")

### DIFF
--- a/src/crates/ai-adapters/src/client.rs
+++ b/src/crates/ai-adapters/src/client.rs
@@ -15,6 +15,7 @@ pub(crate) mod utils;
 use crate::providers::{anthropic, gemini, openai};
 use crate::types::ProxyConfig;
 use crate::types::*;
+use crate::WireReasoningFields;
 use anyhow::Result;
 use format::ApiFormat;
 use log::warn;
@@ -31,6 +32,8 @@ pub struct StreamResponse {
         Box<dyn futures::Stream<Item = Result<crate::stream::UnifiedResponse>> + Send>,
     >,
     pub raw_sse_rx: Option<mpsc::UnboundedReceiver<String>>,
+    /// Thinking / effort values extracted from the final outbound request body.
+    pub wire_reasoning: WireReasoningFields,
 }
 
 /// Default time to wait for the first response headers / stream body to start.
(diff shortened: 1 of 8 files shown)


limityan (changes requested):
Thanks for the PR. The product problem is real: users and support need a reliable way to verify which reasoning/thinking settings actually reached the provider after adapter defaults, normalization, and request-body overrides. Logging the values from the final outbound request body is the right direction for diagnostics, and it is better than logging only the user-facing config value.

I would treat this as an internal/support observability feature rather than a user-facing product feature. It lowers debugging and support cost, but users will still need a UI or diagnostics surface if the goal is for them to understand reasoning behavior without digging through `ai.log`.

Blocking issues before merge:

1. The PR is not based on the current `GCWing/main`. Current main is `9fa991f6`, while the PR merge ref is still based on `be14b80e`. Please rebase/update the branch and rerun the relevant checks before merging.

2. The stated coverage is not complete for Gemini Code Assist. That adapter wraps the real Gemini request under a top-level `request` field, while `wire_reasoning.rs` only inspects top-level fields such as `/generationConfig/...`, `/reasoning/...`, and `/output_config/...`. As a result, Code Assist can have thinking enabled in the actual outbound payload but still log `wire_thinking=None` / `wire_effort=None`. Please either unwrap the `request` envelope in the extractor or explicitly support `/request/...` paths, and add a regression test with a wrapped Code Assist request body.

Verification performed locally on the PR merge ref:

- `git diff --check` for the final 6-file PR diff passed.
- `cargo test -p bitfun-ai-adapters wire_reasoning -- --nocapture` passed.
- `cargo check -p bitfun-core` passed.
- `cargo fmt --all --check` reports pre-existing repository-wide formatting drift, but the touched Rust files should still be formatted before final CI if formatting is enforced.

Once the Code Assist coverage hole is fixed and the branch is rebased onto current main, I think the product direction is reasonable to accept as diagnostic observability.

PR #955, opened 2026-05-29
[codex] Dedupe Rust dependency versions

limityan:
## Summary

- Move the workspace `dashmap` requirement to 6.x so direct BitFun crates align with the `russh-sftp` graph.
- Use the workspace `similar` requirement from the CLI crate instead of a separate broad requirement.
- Move relay-server's direct `dashmap` requirement onto the workspace dependency.

## Impact

This removes `dashmap v5.5.3` from the Rust dependency graph and keeps the remaining direct `similar` requirements aligned at the workspace level.

## Validation

- Confirmed `dashmap@5.5.3` is no longer present in `cargo tree --workspace --target all`.
- Ran `cargo check --workspace` successfully. The only warning observed was the existing desktop dead-code warning for `parse_clipboard_path_segments`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -68,7 +68,7 @@ regex = "1"
 base64 = "0.22"
 image = { version = "0.25", default-features = false, features = ["png", "jpeg", "gif", "webp", "bmp"] }
 md5 = "0.7"
-dashmap = "5"
+dashmap = "6"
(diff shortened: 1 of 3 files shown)


limityan:
Closing this split PR because the dependency dedupe and modernization commits have been consolidated into #954.

PR #1309, opened 2026-06-25
fix(prompt): discourage small-limit probe reads in Read tool description

nonoqing:
## Summary

- Models occasionally call `Read` with `limit: 5` to "probe" a file's type before reading it properly
- Source files begin with copyright headers, so probe reads return no useful code and waste a round trip
- The existing anti-fragmentation guidance targets 30-100 line chunks (progressive slicing) but misses this distinct pattern
- Add an explicit bullet calling out small `limit` values (< 50) as a probe anti-pattern

## Test plan

- [ ] Observe Read tool calls in eval traces — `limit: 5` probe pattern should no longer appear
- [ ] Verify legitimate range reads (e.g. `start_line: 200, limit: 50`) are unaffected

🤖 Generated with [Claude Code](https://claude.com/claude-code)


nonoqing:
closing: will re-submit from nonoqing:evals-on-release instead

PR #1658, opened 2026-07-21
feat(external-sources): improve subagent detail, OpenCode card, MCP empty state

limityan:
## Summary

Improve the External AI applications settings UI while keeping external-source identity and host paths on their owning side of the boundary:

- make subagent details and source locations easier to inspect
- aggregate OpenCode configuration with lifecycle, diagnostics, scope, and capability controls intact
- explain the MCP empty state only after a successful snapshot
- reveal local source locations without sending raw or redacted display paths back from the frontend

### Safe source-location actions

- The frontend sends only `workspacePath` plus the opaque source `stableKey`.
- The core resolves that identity against its private raw discovery snapshots; raw host paths remain absent from the public snapshot.
- The desktop adapter owns the actual file-manager reveal operation.
- Remote workspaces declare this command unsupported, and the UI renders location text as disabled there.
- Repeated redacted labels no longer merge distinct Subagent or OpenCode sources.

### OpenCode and source status

- Aggregate OpenCode sources into the dedicated card without dropping per-source lifecycle state, enable controls, or diagnostics.
- Show all represented scopes and include tool counts in the summary.
- Keep independently identified configuration locations separate even when their redacted labels match.
- Reuse the same source-member control renderer for OpenCode and other ecosystems.

### Empty, error, and presentation states

- Show MCP discovery guidance only when a snapshot loaded successfully and contains no MCP servers, approvals, or conflicts.
- Replace retired/undefined CSS variables with current theme tokens.
- Keep Subagent source locations collapsible and diagnostics localized.

## Compatibility and architecture impact

- Adds one narrow core host-action API for stable-key-to-local-path resolution and registers it in the core public API boundary rules.
- Adds one desktop Tauri command with an explicit `RemoteUnsupported` policy.
- No raw path is added to serialized DTOs, and no remote execution behavior changes.

## Verification

- `pnpm --dir src/web-ui run test:run src/infrastructure/api/service-api/ExternalSourcesAPI.test.ts src/infrastructure/config/externalSourcePresentation.test.ts src/infrastructure/config/components/ExternalSourcesConfig.test.tsx` — PASS (64/64)
- `pnpm run type-check:web` — PASS
- `pnpm run lint:web` — PASS (0 errors; pre-existing warnings only)
- `pnpm run i18n:audit` — PASS (0 warnings)
- `pnpm run theme:color-audit:all` — PASS
- `node scripts/check-core-boundaries.mjs` — PASS
- `pnpm run check:repo-hygiene` — PASS
- `pnpm run check:github-config` — PASS (7/7)
- `cargo check --workspace` — PASS
- `cargo test -p bitfun-core reveal_location_uses_stable_identity_and_keeps_the_raw_path_private` — PASS
- `cargo test -p bitfun-desktop remote_workspace_policy::tests` — PASS (2/2)
- `git diff --check gcwing/main..HEAD` — PASS

### DIFF
--- a/scripts/core-boundaries/rules/source/public-api-rules.mjs
+++ b/scripts/core-boundaries/rules/source/public-api-rules.mjs
@@ -576,6 +576,12 @@ export const externalSourceCorePublicApiEntries = [
       'BitFun CLI and desktop host APIs',
     ),
   ),
+  externalSourceEntry(
+    'external_source_location_for_host_action',
+    'bitfun-core external source composition owner',
+    'Desktop external-source configuration host adapter',
+    true,
+  ),
   ...[
     'ExternalToolActivationState',
     'ExternalToolApprovalRequest',
--- a/src/apps/desktop/src/api/commands.rs
+++ b/src/apps/desktop/src/api/commands.rs
@@ -3532,8 +3532,15 @@ pub async fn reveal_in_explorer(
             ))
         }
     };
+    reveal_local_path_in_explorer(path, &request.path)
+}
+
+pub(crate) fn reveal_local_path_in_explorer(
+    path: &std::path::Path,
+    display_path: &str,
+) -> Result<(), String> {
     if !path.exists() {
-        return Err(format!("Path does not exist: {}", request.path));
+        return Err(format!("Path does not exist: {display_path}"));
     }
     let is_directory = path.is_dir();
     let path_str = path.to_string_lossy().to_string();
--- a/src/apps/desktop/src/api/external_sources_api.rs
+++ b/src/apps/desktop/src/api/external_sources_api.rs
@@ -1,7 +1,8 @@
 //! Desktop host API for ecosystem-neutral external AI application sources.
 
 use bitfun_core::external_sources::{
-    choose_external_mcp_conflict, choose_external_subagent_conflict, external_source_snapshot,
+    choose_external_mcp_conflict, choose_external_subagent_conflict,
+    external_source_location_for_host_action, external_source_snapshot,
     set_external_mcp_server_decision, set_external_prompt_command_conflict_choice,
     set_external_source_enabled, set_external_subagent_activation,
     set_external_tool_conflict_choice, set_external_tool_target_decision,
@@ -28,6 +29,13 @@ pub struct SetExternalSourceEnabledRequest {
     pub source_key: String,
     pub enabled: bool,
     pub expected_preference_revision: u64,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+pub struct RevealExternalSourceLocationRequest {
+    pub workspace_path: Option<String>,
+    pub source_key: String,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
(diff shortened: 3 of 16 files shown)


limityan (commented):
## Code Review: PR #1658 — External sources config UI improvements

Reviewed all 6 files. Overall the UI improvements are well-structured — the OpenCode aggregation card, MCP empty state, and simplified subagent detail all address real UX issues. The i18n keys are consistent across en-US / zh-CN / zh-TW with matching placeholders. However, there is one logic bug and several issues worth addressing before merge.

---

### Bug: countKey mapping for `mcp` and `tool` is incorrect

**File:** `src/web-ui/src/infrastructure/config/components/ExternalSourcesConfig.tsx` (lines ~1199-1201)

```tsx
const countKey = capabilityId === 'subagent' ? 'agents'
  : capabilityId === 'command' ? 'commands'
  : capabilityId;
```

The `ExternalSourceCapabilityCounts` type (in `externalSourcePresentation.ts`) has keys: `commands`, `tools`, `agents`, `mcps` — all plural. The mapping correctly converts `subagent` → `agents` and `command` → `commands`, but leaves `mcp` and `tool` as-is (singular). Since `g.counts['mcp']` and `g.counts['tool']` are `undefined`, the `?? 0` fallback always yields **0**.

**Impact:** The per-capability description for MCP always shows "0 MCP servers found" and tool always shows "0 tools found", even when servers/tools exist. Note the summary line at the top of the card correctly uses `g.counts.mcps` (line ~1140), creating an inconsistency where the summary shows non-zero counts but the expanded capability rows show 0.

**Fix:**
```tsx
const countKey = capabilityId === 'subagent' ? 'agents'
  : capabilityId === 'command' ? 'commands'
  : capabilityId === 'mcp' ? 'mcps'
  : capabilityId === 'tool' ? 'tools'
  : capabilityId;
```

---

### Accessibility: `<a>` without `href` is not keyboard-accessible

**File:** `ExternalSourcesConfig.tsx`, `renderPathLink` (lines ~907-930)

```tsx
<a
  className="bitfun-external-sources-config__path-link"
  title={location}
  translate="no"
  onClick={(event) => {
    event.preventDefault();
    void workspaceAPI.revealInExplorer(location).catch(() => {});
  }}
>
  {display}
</a>
```

An `<a>` without `href` is not focusable, not keyboard-activatable, and screen readers may not announce it as a link. The `event.preventDefault()` is also a no-op without `href`.

**Recommendation:** Add `href="#"` (and keep `preventDefault`), or use a `<button>` styled as a link, or add `role="button"` + `tabIndex={0}` + `onKeyDown` handler for Enter/Space.

---

### Design concern: Fragile string-based OpenCode detection for source groups

**File:** `ExternalSourcesConfig.tsx` (lines ~458-463)

```tsx
const opencodeGroups = useMemo(
  () => sourceGroups.filter((group) => group.displayName.toLowerCase().includes('opencode')),
  [sourceGroups],
);
```

This is inconsistent with the ecosystem-level detection which uses the robust `ecosystemId === 'opencode'` check (line ~1120). If source display names ever change (the test data in this PR already changed them from "OpenCode project commands" to "External project commands"), `opencodeGroups` would silently become empty — the card would show 0 counts and no config locations.

**Recommendation:** Consider adding `ecosystemId` to `ExternalSourcePresentationGroup` and filtering on that instead of string-matching on `displayName`.

---

### Design concern: Hardcoded `scope.user_global` in capability descriptions

**File:** `ExternalSourcesConfig.tsx` (line ~1223)

```tsx
{t(`opencode.capability.${capabilityId}.description`, {
  count,
  scope: t('scope.user_global'),
})}
```

The `scope` parameter is always `t('scope.user_global')` regardless of the actual source scope. For project-scoped OpenCode sources, the description would still say "from User global", which is misleading.

---

### Test coverage gap: OpenCode aggregation card is untested

**File:** `ExternalSourcesConfig.test.tsx`

The test data was changed to use "External project commands" / "External user configuration" instead of "OpenCode ...", which means `opencodeGroups` is empty in all tests (the filter checks for `includes('opencode')`). No test covers the OpenCode aggregation card with actual data — the mode selector, capability switches, config locations, and summary counts are all untested.

**Recommendation:** Add a test case with display names containing "OpenCode" to verify the aggregation card renders correctly.

---

### Minor issues

1. **String literal instead of enum** (line ~742): `workspace?.workspaceKind === 'remote'` uses a string literal instead of the imported `WorkspaceKind.Remote` enum value.

2. **Unused i18n key**: The old `agents.sources` key ("Combined from {{count}} configuration sources") was replaced by `agents.sourceLocations` but was not removed from the locale files. Consider removing it to avoid dead i18n entries.

3. **Silent error swallowing** (line ~928): `void workspaceAPI.revealInExplorer(location).catch(() => {})` silently discards all errors. This is acceptable for a convenience action, but consider at least logging the error for debugging purposes.

---

### Summary

The `countKey` mapping bug should be fixed before merge — it produces incorrect UI for MCP and tool capability descriptions. The accessibility issue with the `<a>` without `href` should also be addressed. The fragile OpenCode detection and missing test coverage are worth addressing but are lower priority.

PR #1757, opened 2026-07-25
fix: resolve 36 findings from the Jul 24 upstream review

bobleer:
Follow-ups from a review of the 16 commits merged to `main` on Jul 24 (#1727,
#1728, #1736, #1739, #1742, #1743, #1746, #1748, #1750, #1751, #1752, #1754).

44 findings were raised; each was re-checked against the tree at `21c0382d4`
before anything was changed. **41 confirmed, 3 were not defects as described.**
This PR fixes 36 of them. Six areas are touched, so there is one commit per area
— reviewing commit-by-commit will be much easier than reading the combined diff.
Contributing guidance asks for focused PRs; these are independent fixes bundled
only because they came out of one review pass, and I am happy to split them.

### The ones worth reading first

**Zombie SSH sessions (`fix(remote-ssh)`)** — dropping a russh `Handle` closes
the socket without telling the peer, so the server keeps the session until its
own timeout. With ProxyJump that is N+1 abandoned sessions per disconnect, one
per hop, accumulating against `MaxSessions`. Two of the six leak sites build
sessions that are never registered anywhere, so nothing else could ever have
closed them — the connection dialog's Test button leaked a session per hop, per
click.

**Reconnect stampede (`fix(remote-ssh)`)** — worse than the review described. A
successful reconnect installs a *new* `alive` Arc, so the flag captured before
taking the lock stays false forever and *every* queued waiter reconnects again.
The lock exists to prevent exactly that. Fixed by moving the locks out of the
map entry that gets replaced, and re-reading liveness under the lock.

**China detection overriding a definite answer (`fix(relay)`)** — a resolved
country of US/DE/JP fell through to a connectivity probe that reports "CN"
whenever `mirrors.aliyun.com` is reachable and `github.com` is not. Aliyun
answers from anywhere; GitHub has outages, rate limits and corporate blocks. One
transient GitHub failure was enough to rewrite a Frankfurt VPS's apt sources,
point Docker at a Chinese registry and route GitHub clones through a third-party
proxy.

**Unverified root code execution (`fix(relay)`)** — in CN mode the Docker install
script defaulted to a jsDelivr copy of `docker/docker-install@master`: a floating
ref, edge-cached for hours, over a third-party CDN, executed as root with no
checksum or signature. Now requires a pinned digest or byte-identical content
from a second independent origin, failing closed.

**apt over https before `ca-certificates` exists (`fix(relay)`)** — mirror init
runs before anything is installed, and minimal cloud images use `http://`
sources precisely because there is no CA bundle. The rewrite broke the very
`apt-get update` that would have installed one, with a TLS error that says
nothing about mirrors.

**Unbounded download into memory (`fix(release)`)** — the checksum can only be
verified once the whole body has arrived, so a hostile origin can stream until
the process is OOM-killed with no integrity check ever running.

**Mirrored assets were never checksummed (`fix(release)`)** — the sync script
copies an archive and its `.sha256` from GitHub without comparing them. Clients
verify an archive against a sidecar from the same host, which proves nothing if
both were copied from a bad download.

The remaining commits cover a `elementFromPoint` layout thrash on every scroll
event, a one-frame glow flicker, a modal scroll lock that unlocked while another
modal was open, a search loop burning ~22 redundant layout passes, and a
"tool not installed" message shown for `tar: link: Not found in archive`.

### Deliberately not changed

- **`recursion_limit = "256"`** — measured rather than assumed. At 128 the build
  fails on `Vec<slab::Entry<h2::…recv::Event>>: Send`, ~15 frames of which are
  inside h2 itself. `Box::pin` cannot collapse third-party depth; only erasing a
  mid-chain future to `Pin<Box<dyn Future + Send>>` would, at the cost of an
  allocation and dynamic dispatch on the dialog-turn path. Rationale recorded in
  `lib.rs`.
- **The light-theme glow filter chain** — `saturate`/`brightness` *are* the look
  #1751 shipped; the accent mix resolves to a blue that the filters push to a
  bright cyan no `color-mix` of the token reproduces. Removing them is a design
  call, not a bug fix. Only the provably invisible `hue-rotate(4deg)` was
  dropped.
- **`docker exec -it <c> <shell> -lc <shell>`** is not a nested double shell.
  Shells `exec` the final simple command of `-c`, so one process lands on the
  PTY, with the login profile sourced — which a bare `docker exec -it … sh`
  would skip. Documented at the call site instead.

### Compatibility

No persisted format changed: no config schema, no manifest field, no Tauri
command signature, no on-disk state layout. Mixed-version interop was checked
for the update lock (old `fs::write` vs. new `create_new` — compatible via the
shared staleness rule) and for the background-install marker (an older child
ignores the new env var).

Six hazards in the first cut of these fixes were caught and corrected before
this PR, most of them worth naming because they are the kind that only bite in
production:

- Tagging a failed hop with `anyhow`'s `.context()` **replaces** the error's
  `Display`, which would have turned every user-visible SSH failure into
  "connection stage 'target'". Replaced with a transparent wrapper; a test pins
  `{}`, `{:#}` and the tag together.
- An age-based sweep of pid files would delete the pid file of a supervised
  command still running after an hour — that file is how cancellation finds the
  process. Now liveness-based (`kill -0`).
- Automatic reconnect on config drift would have killed an in-flight exec or PTY
  on a still-healthy session. Stale handles are only closed when nothing else
  holds the `Arc`; an explicit `disconnect()` keeps no such guard.
- `disconnect_all()` now does network sends, so it is bounded at 1s per handle —
  it runs on app shutdown, and the session loop may be wedged on an unwritable
  socket.
- On re-deploy the new apt path would have overwritten the saved upstream
  `sources.list` with already-mirrored content, leaving `restore_apt` nothing
  real to restore.
- Checksum verification globbing `*.sha256` also matched Desktop sidecars, whose
  format that script does not own; scoped to the manifest's own asset list.

Path rejection was narrowed to `..` only — `.` is not traversal and `./name` has
always worked — and `restart_managed_daemon` checks both `dirs::config_dir()`
and `~/.config` so a unit from an older install is still found.

### Testing — fully tested (automated); UI changes not visually verified

`cargo check` and `clippy` clean on every touched crate. 386 integrations lib
tests, 217 desktop tests, 10 CLI self-update tests, 2048 web-ui tests, 37 i18n
contract tests, `tsc --noEmit` and eslint all pass. New tests cover the stage
tag's message preservation, the pid sweep sparing a live process, the narrowed
tool-missing detection, remote path traversal, and the prevent-sleep rollback
and `ConfigReloaded` paths.

`tests/file_watch_contracts.rs` has 3 failures on my machine that are
pre-existing — confirmed identical against a clean stash of `main`.

Two changes alter appearance and I have **not** verified them visually, so they
deserve a look: dropping `hue-rotate(4deg)` from the light-theme glow, and
removing `!important` from the interactive-surface `transition-property`. The
latter means components declaring their own `transition` now fully own it, which
is the contract the zero-specificity `:where()` default should have had, but it
can change hover animation on buttons that were relying on the override.

AI-assisted (Claude Code); every finding was re-verified against the code before
being acted on, and the three that did not hold up are called out above rather
than "fixed".

### DIFF
--- a/scripts/openbitfun-release-sync.sh
+++ b/scripts/openbitfun-release-sync.sh
@@ -89,6 +89,46 @@ download_asset() {
   fi
 }
 
+# Check the mirrored Linux archives against the `.sha256` sidecars mirrored with
+# them. Reads the filename list on stdin, one per line.
+#
+# Scoped to the manifest's own assets rather than globbing `*.sha256`: the same
+# directory also holds the Desktop packages, whose sidecars this script does not
+# own and must not assume the format of. Deleting a good archive because a
+# foreign sidecar was laid out differently would be worse than not checking.
+#
+# A mismatch means this run pulled a bad copy from GitHub, so the archive is
+# removed: `download_asset` skips files that already exist, and leaving a corrupt
+# one in place would make every later run treat it as done. Removing it lets the
+# next run re-fetch and self-heal.
+verify_mirrored_checksums() {
+  local filename sidecar archive expected actual failed=0
+  while IFS= read -r filename; do
+    [ -n "$filename" ] || continue
+    case "$filename" in
+      *.sha256) ;;
+      *) continue ;;
+    esac
+    sidecar="${VERSION_DIR}/${filename}"
+    archive="${sidecar%.sha256}"
+    if [ ! -f "$sidecar" ] || [ ! -f "$archive" ]; then
+      continue
+    fi
+    expected="$(awk '{print $1; exit}' "$sidecar")"
+    actual="$(sha256sum "$archive" | awk '{print $1}')"
+    if [ "$expected" != "$actual" ]; then
+      log "ERROR: checksum mismatch for $(basename "$archive"): expected $expected, got $actual"
+      rm -f "$archive"
+      failed=1
+    fi
+  done
+  if [ "$failed" -ne 0 ]; then
+    log "ERROR: refusing to publish a manifest for unverified assets"
+    return 1
+  fi
+  return 0
+}
+
 # Fetch linux-binaries.json into $LINUX_MANIFEST_TMP, retrying transient
 # failures. Sets LINUX_MANIFEST_STATE to one of:
 #   ok        — downloaded
@@ -165,6 +205,16 @@ PY
       log "  Mirroring Linux binary asset: $filename"
       download_asset "$url" "${VERSION_DIR}/${filename}" || exit 1
     done <<< "$LINUX_ASSET_LIST"
+
+    # Verify before publishing the manifest that points at these bytes.
+    #
+    # Clients check an archive against the checksum served next to it. When both
+    # come from here, mirroring a corrupted archive alongside its original
+    # checksum is the one failure that verification cannot catch — every
+    # downstream install would then fail, or worse, succeed on bad bytes if the
+    # corruption also reached the sidecar. Checking here is the only place the
+    # two copies can still be compared.
+    cut -f2 <<< "$LINUX_ASSET_LIST" | verify_mirrored_checksums || exit 1
 
     "$PYTHON" - "$LINUX_MANIFEST_TMP" "${VERSION_DIR}/linux-binaries.json" \
       "$OPENBITFUN_BASE_URL" <<'PY'
--- a/src/apps/cli/src/self_update.rs
+++ b/src/apps/cli/src/self_update.rs
@@ -40,6 +40,14 @@ const READ_TIMEOUT: Duration = Duration::from_secs(30);
 const MANIFEST_TIMEOUT: Duration = Duration::from_secs(20);
 /// Ceiling for the whole startup-path check, which precedes the first paint.
 const AUTO_CHECK_BUDGET: Duration = Duration::from_secs(10);
+/// Hard ceiling on an archive held in memory. The checksum can only be verified
+/// once the whole body has arrived, so without a cap a hostile or misconfigured
+/// origin can stream until the process is OOM-killed and no integrity check ever
+/// runs. Official CLI archives are tens of megabytes; this leaves ample room.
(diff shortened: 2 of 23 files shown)


bobleer:
CI's `CLI Tests (ubuntu-latest)` job failed on the first run. Investigated — it is **not** caused by the changes in this PR, but I pushed a fix since it was blocking:

```
thread 'legacy_reports_a_missing_primary_without_recursing' panicked at src/apps/cli/tests/compat_entrypoint.rs:52:
run isolated deprecated launcher: Os { code: 26, kind: ExecutableFileBusy, message: "Text file busy" }
```

`compat_entrypoint.rs` is untouched by this PR — the only `bitfun-cli` file it changes is `self_update.rs`, and neither path reaches the other: the legacy launcher just `exec`s its sibling and never calls into `self_update`, and `--version` exits before the auto-update hook. The full `-p bitfun-cli` suite (340 unit + all integration tests) passes locally.

The cause is a Linux-only race between the file's three tests, which run as threads of one process. `legacy_reports_a_missing_primary_without_recursing` copies the launcher into a tempdir and execs it; while `fs::copy` holds the destination open for writing, a sibling thread forking for its own `Command` hands the child an inherited write descriptor, and `execve` of that path returns `ETXTBSY` until the child clears it. The descriptor is `CLOEXEC`, so the window is tiny — which is why it is green almost always and lost the coin toss here.

`test(cli): retry the compat launcher exec past a Linux ETXTBSY race` retries the exec. The writer is already closed by the time `fs::copy` returns and the inherited copy is out of the caller's hands, so retrying is the fix available at this layer. Happy to split it into its own PR if you would rather keep this one to the review findings.

Separately, and left alone as out of scope: `cargo clippy -p bitfun-cli --all-targets` fails on `services-core/src/json_store.rs:335` (`create` without `truncate`). Reproduced on `main` with this branch stashed, and CI does not run clippy, so it is neither new nor blocking.

PR #1779, opened 2026-07-26
refactor(browser-control): consolidate to a single default browser path

bobleer:
## Background

Follow-up to #1777. The audit identified seven overlapping browser-control paths as the main reason the agent performs poorly at browser operations: the model oscillates between two full automation stacks (the agent-browser CLI skill and the ControlHub CDP domain), each with its own `@eN` ref namespace, its own browser instance, and mutually invisible login state — refs from one stack fail in the other, and recovery guidance pointed in both directions.

This PR makes the ControlHub browser domain the single default path and fixes the correctness issues that remained on it.

## Changes

**Single default stack**
- The agent-browser builtin skill is now default-off in every mode (`DISABLE_COMPUTER_USE` rule added to `AGENTIC_POLICY`/`TEAM_POLICY`). Both opt-in paths are preserved and contract-tested: user overrides, and explicit `/skill agent-browser` invocation (the default-hidden whitelist now includes the `computer-use` group — without this, explicit invocation would have broken).

**Interaction correctness on the ControlHub path**
- iframe click fix: `element_center` accumulates ancestor frame offsets (rect + border) along the `frameElement` chain, so clicks inside same-origin iframes land on the element instead of iframe-local coordinates. Cross-origin frames return a structured error advising a fresh snapshot and re-target; missing coordinates now error instead of silently clicking (0,0). All element-targeting actions (`click`, `hover`, `fill`, …) benefit.
- Removed the dead `frame`/`frame_main` actions: they wrote `active_frame` state that nothing ever read, so the model believed it had switched frames when nothing changed. The unread session state is deleted too.

**Recoverable errors**
- High-frequency failures now carry stable error codes and recovery instructions at the source: element not found → `NOT_FOUND` ("take a new snapshot and use a fresh @eN ref"), dead session/WebSocket → `WRONG_TAB` ("call browser.connect or switch_page"), CDP/wait timeouts → `TIMEOUT`. The phrase-matching classifier in `map_dispatch_error` remains as fallback only.

**Honest observability at path boundaries**
- `open_builtin` now states the built-in panel is display-only and not observable by the agent, routing reading/interaction to `connect` + `snapshot`.
- `open_url`/`open_file` results explain the opened window cannot be observed or controlled by BitFun and route to the browser domain / desktop actions respectively — previously the model would open a URL and then try to drive the browser with desktop clicks.
- The WebFetch description routes login/JS-dependent pages to the browser domain.

**Plan update**
- Records the product decision that `ai.computer_use_enabled` and browser control are independent switches: disabling computer use must not hide or disable the ControlHub browser domain; phase 6 will add a separate `ai.browser_control_enabled` (default on) plus a `browser_control` permission intent.

## Verification

- `cargo check --workspace` passes (verified with pipefail; includes `bitfun-desktop`)
- `cargo test -p bitfun-core browser`: 27 passed (new iframe/structured-error/open_builtin tests included)
- `cargo test -p bitfun-core control_hub computer_use skills`: 263+ passed
- `cargo test -p bitfun-agent-runtime` (incl. `skill_contracts`): all passed, with new contract tests for default-off in every mode and explicit-invocation opt-in
- `node scripts/core-boundaries/checker.mjs` passes


bobleer:
## Update: end-to-end review pass

After the consolidation commit I ran an adversarial end-to-end review of the resulting path — four reviewers (task flow, prompts/agent UX, diff correctness, capability gating), each finding independently verified against the code before being accepted. 33 findings survived verification; the fixes are in `fix(browser-control): make browser results actionable and app detection safe`.

Two of them explain why browser automation failed in practice regardless of which stack was used:

**The `@eN` workflow was never executable.** `convert_tool_result` prefers `result_for_assistant` when a tool supplies one, and every ControlHub browser action supplied only a one-line count. A model calling `snapshot` received the literal string `"Snapshot: 42 interactive elements"` — the refs that `connect → snapshot → click @e1` depends on stayed in the data field and never reached it. The same applied to `list_pages`/`tab_query` (no page ids → `switch_page` unusable), `get_html`, `cookies`, `network`/`console`/`errors` and `meta.capabilities`; `screenshot` returned a byte count and no image at all. Every one of these now returns what the next decision needs, and `screenshot` returns a real image attachment.

**`snapshot` could make `click` act on the wrong element while reporting success.** Refs were renumbered from 1 on each pass without clearing the previous pass's `data-cdp-ref` attributes, and numbering is viewport-filtered — so after an `auto_scroll`, `click @e2` could resolve to a stale attribute earlier in document order. Refs are now cleared across the document, open shadow roots and same-origin frames before renumbering.

Also fixed, and worth calling out because it is a regression from #1777: on Windows the foreground app `name` is the window *title*, and the Chromium check used bare substring matching — so any window titled "Search" matched `"arc"` (s-e-**arc**-h) and had all desktop input blocked by the newly-activated guard. Foreground apps now carry a `process_name` and matching is whole-identity.

The rest: `click`/`hover` hit-test their target point instead of reporting success on an occluded element; `press_key` sends `text`/`windowsVirtualKeyCode` so Enter submits forms; `browser.fetch` no longer reports CORS failures as successful fetches; `select`/`wait` accept `@eN` refs; stale refs report `STALE_REF`; cross-origin iframes are explained rather than retried forever; prompts no longer promise the user's login state (connect uses BitFun's managed profile) and no longer point unconditionally at the default-off ComputerUse tool.

Verification: `cargo check --workspace`, `cargo test -p bitfun-core` (1494 passed), `-p bitfun-agent-runtime`, `-p bitfun-agent-tools` (103), `bitfun-desktop` builds, `node scripts/core-boundaries/checker.mjs`.

Known remaining (not in this PR): a guarded desktop action still queries the frontmost app twice per call (guard + result augmentation); caching it within a call needs plumbing the snapshot through and is better done with the per-session host-state work in phase 3 of `docs/plans/computer-use-refactor-plan.md`.

PR #1820, opened 2026-07-28
fix(worktree): defer isolation until first prompt

bobleer:
## Summary

- make the Desktop checkbox and CLI `/worktree` command arm an empty-session preference without running Git I/O
- materialize or release the worktree only after the first prompt is visibly submitted, then start the backend turn in the resolved execution root
- keep the stable project root as the session persistence locator while the worktree remains the execution root, including legacy caller normalization
- expose pending worktree state in both the Desktop branch strip and CLI workspace status

## Type and Areas

Type: regression fix / UI/UX

Areas: Rust core, desktop/Tauri, Web UI, CLI/TUI, Peer Host adapter

## Motivation / Impact

Checking worktree isolation previously performed synchronous worktree creation. The following prompt then passed the worktree execution path as a session-storage locator, causing `Session ID is already bound to another workspace` and preventing the message from being sent.

After this change, checking the control is immediate. Worktree creation starts only after prompt submission, and session persistence continues to resolve through the owning project root on Windows, macOS, and Linux. Remote-workspace support remains explicitly unchanged.

## Verification

- `pnpm run type-check:web`
- `pnpm --dir src/web-ui run test:run src/flow_chat/utils/sessionWorktree.test.ts src/flow_chat/components/ChatInputWorkspaceStrip.test.tsx src/flow_chat/store/FlowChatStore.test.ts` — 79 passed
- `pnpm run lint:web`
- `pnpm run i18n:audit` — 0 warnings
- `cargo check --workspace`
- `cargo test -p bitfun-core worktree_execution_root_is_a_legacy_alias_for_project_storage --lib`
- `cargo test -p bitfun-core unrelated_workspace_is_not_rewritten_to_the_project_storage_root --lib`
- `cargo test -p bitfun-core dialog_port_preserves_invalid_request_for_wrong_workspace --lib`
- `cargo test -p bitfun-cli` — all unit and integration tests passed
- `cargo test -p bitfun-desktop` — 226 passed, 2 pre-existing permission-gated tests ignored
- `node scripts/check-core-boundaries.mjs`
- `git diff --check`

## Reviewer Notes

The storage-locator fallback intentionally rewrites only the loaded session own execution root to its project root. Unrelated workspace paths are preserved and still rejected by the existing session identity guard. No screenshot is included because the UI shape is unchanged; the visible behavior change is that the checked state is immediate and reports a pending state until first send.

This change was AI-assisted and fully tested with the commands above.

## Checklist

- [x] User-facing strings, docs, and locales are updated where applicable.

### DIFF
--- a/src/apps/cli/src/agent/runtime_client.rs
+++ b/src/apps/cli/src/agent/runtime_client.rs
@@ -822,7 +822,9 @@ impl CliAgentRuntimeClient {
             original_message: None,
             turn_id: Some(turn_id.clone()),
             agent_type: agent_type.to_string(),
-            workspace_path: Some(self.workspace_path_string()),
+            // Dialog submission uses this path to locate persisted session
+            // state. Execution still comes from the session's resolved binding.
+            workspace_path: Some(self.project_workspace_path_string()),
             remote_connection_id: None,
             remote_ssh_host: None,
             policy: DialogSubmissionPolicy::for_source(AgentSubmissionSource::Cli),
--- a/src/apps/cli/src/chat_state.rs
+++ b/src/apps/cli/src/chat_state.rs
@@ -315,6 +315,9 @@ pub(crate) struct ChatState {
     is_git_repository: bool,
     /// Whether this runtime exposes session worktree lifecycle controls.
     worktree_control_available: bool,
+    /// Empty-session preference. The actual worktree is created only after the
+    /// user submits the first prompt.
+    worktree_isolation_requested: Option<bool>,
     /// Current model display name (shown in shortcuts bar)
     pub current_model_name: String,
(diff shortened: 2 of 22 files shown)


bobleer:
CI follow-up: the first Ubuntu full-core run exposed that an omitted workspace locator was being synthesized from the loaded session and re-resolved against ambient storage settings. Commit `3899c804b` now preserves the loaded session storage binding when the caller omits the locator, while still normalizing an explicitly supplied worktree execution root to the project root.

Reverification after the fix:
- both originally failing scheduler tests pass
- the new omitted-locator regression test passes
- explicit worktree-root normalization and wrong-workspace rejection tests pass
- `cargo test -p bitfun-core --lib`: 1543 passed, 1 ignored

PR #2081, opened 2026-08-05
fix(flow-chat): refresh context usage display right after compression completes

YodonTan:
## Problem

After a context compression completes, the input-box context usage percentage and the "last request input context" tooltip (e.g. `128.0K/128.1K (100%)`) kept showing the pre-compression values.

Root cause: `session.currentTokenUsage` is only written by the `TokenUsageUpdated` event, which is only emitted after a model response. A completed compression only emits `ContextCompressionCompleted` (which carries `tokens_after`), and the frontend handler only updated the compression tool card without writing the compacted size back to the store. For manual compression there is no following request, so the display stayed stale indefinitely.

## Fix

In `handleCompressionCompleted` (src/web-ui/src/flow_chat/services/flow-chat-manager/EventHandlerModule.ts), when `applied === true` and `tokensAfter` is a valid non-negative number, write it into `session.currentTokenUsage` via `store.updateTokenUsage` (without a `dialogTurnId`, so compression does not accumulate into the turn's token usage).

The existing subscription chain (ChatInput selector fingerprint -> setTokenUsage -> ModelSelector percentage + tooltip) picks the value up immediately, so the input box refreshes right after compression finishes. A later model response still overwrites it with the exact provider-reported usage.

No backend change: re-emitting `TokenUsageUpdated` after compression would be recorded by `TokenUsageSubscriber` as a model usage record and pollute usage statistics.

## Tests

- `EventHandlerModule.test.ts`: 3 new cases (applied -> currentTokenUsage updated; not applied -> unchanged; invalid `tokensAfter` -> unchanged). All 32 tests pass.
- `tsc --noEmit` for web-ui passes.

Verified manually with a built installer: after compression, the displayed token count updates immediately.

### DIFF
--- a/src/web-ui/src/flow_chat/services/flow-chat-manager/EventHandlerModule.test.ts
+++ b/src/web-ui/src/flow_chat/services/flow-chat-manager/EventHandlerModule.test.ts
@@ -15,6 +15,8 @@ import type { DialogTurn, FlowToolItem, FlowUserSteeringItem, ModelRound, Sessio
 import type { FlowChatContext } from './types';
 import { markOptimisticDispatchTurnMetadata } from '@/features/dispatch/optimisticDispatchTurn';
 
+const { handleCompressionCompleted } = __test_only__;
+
 vi.mock('../../../shared/notification-system/services/NotificationService', () => ({
   notificationService: {
     error: vi.fn(),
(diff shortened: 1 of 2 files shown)


YodonTan:
Hi maintainers, could you please approve / re-trigger the CI for this PR?

The `CLI Tests (macos-15)` failure on the previous run is **not caused by this PR**:

- This PR only touches 2 frontend files (`src/web-ui/src/flow_chat/services/flow-chat-manager/EventHandlerModule.ts` and its test); the CLI is a separate Rust TUI (`src/apps/cli`) with no dependency on web-ui code.
- The failing test `startup_bracketed_paste_attaches_an_image_path_without_rendering_the_path` is a timing race: it sent Ctrl+C right after the stream-completed marker, before the turn state machine settled back to idle, so the shortcut was interpreted as "interrupt" instead of "exit". This was fixed upstream in `0015d3fdf` ("test(cli): wait for idle before sending exit shortcut"), which is already in `main`.
- I rebased this branch onto the latest `main` (`adb47a171`), so the run now includes that fix. The PR diff is still only the 2 web-ui files (91 insertions).

However, the new push created a workflow run that GitHub marked as `action_required`, which needs a maintainer approval (the branch is from a fork). Could you click **Approve and run** (or re-run the workflow) on the latest run? Thanks!

YodonTan:
Tracking issue for the bug fixed by this PR (context usage display not refreshing after compression): #2094

limityan (approved):
审核结论：通过。已检查压缩完成事件契约、前端 store 更新逻辑、ChatInput 刷新链路及新增测试。`applied === true` 且 `tokensAfter` 有效时更新 session 当前上下文用量，不传 dialogTurnId，避免污染本轮模型 token 累计；未发现需要阻止合并的功能性问题。建议后续在 CompressionEvent 类型中补充 `applied?: boolean`，但不影响本 PR。请等待剩余 CI 检查全部通过后合并。

PR #2086, opened 2026-08-05
fix(runtime): register session mode port on desktop session surface

YodonTan:
## Problem

Switching any Session from the default Agent mode to another mode (e.g. Agent -> Plan) in the desktop app always failed with "切换智能体模式失败，请重试" (Failed to switch agent mode, please retry).

## Root cause

The desktop session surface runtime (`CoreServiceAgentRuntime::session_surface_agent_runtime`, the only runtime surface used by the desktop app) never registered the `session_mode` port. Every `update_session_mode` call therefore hit the missing-port guard in `AgentRuntime::update_session_mode` and returned `PortErrorKind::NotAvailable` with "agent session mode port is not registered". This affected all sessions and all target modes. Model switching was unaffected because `session_model` was registered.

The port registration was missing since the desktop session lifecycle convergence (`f98306168d`).

## Fix

- Register the coordinator-backed `AgentSessionModePort` in `session_surface_agent_runtime`.
- Add a source contract test (`session_surface_runtime_registers_session_mode_port`) so the registration cannot silently regress again.
- Add failure-path diagnostics (desktop command error, coordinator binding failure, registry resolution failure incl. external-route shadowing, persistence failure, frontend log with session target) so regressions are diagnosable from `app.log`.

## Verification

- `cargo check -p bitfun-core --features product-full` and `cargo check -p bitfun-desktop`: pass
- Coordinator session-mode tests, `service_agent_runtime` tests (including the new contract test), and registry external-route tests: all pass
- Frontend type-check and Vite build pass

### DIFF
--- a/src/apps/desktop/src/api/agentic_api.rs
+++ b/src/apps/desktop/src/api/agentic_api.rs
@@ -1723,14 +1723,22 @@ pub async fn update_session_mode(
         request.include_internal,
     )
     .await?;
-    runtime
+    let result = runtime
         .agent_runtime()
         .update_session_mode(AgentSessionModeUpdateRequest {
             session_id,
             mode_id: request.mode_id,
         })
-        .await
-        .map_err(|error| format!("Failed to update session mode: {}", error.into_message()))
+        .await;
+    if let Err(error) = &result {
+        log::error!(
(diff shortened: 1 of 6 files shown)


wsp1911:
#2087

PR #2116, opened 2026-08-06
docs(readme): require reproducible benchmark evidence

bobleer:
## Summary

- replace unsupported homepage performance numbers with implementation-level descriptions
- replace the single-run comparison section with an explicit reproducibility standard
- remove three benchmark images whose raw runs, commands, and pinned environments are not available in the repository

## Why

The current 98.67%, 94.6%, and 36.1x claims cannot be reproduced from public repository evidence. The comparison chart also lacks a dataset snapshot, harness and baseline versions, model settings, raw trajectories, and aggregation method. A caveat does not make those headline numbers verifiable.

The replacement keeps the underlying product facts: BitFun records provider-reported cache reads separately from writes and can use a resident workspace-search index. Numerical results can return once the commands, raw outputs, and pinned environment are published with them.

## Verification

- `git diff --check`
- unsupported-claim/reference search returns no matches in the two READMEs or `png/`


bobleer:
Withdrawing this PR. I reviewed a non-main worktree state and treated the missing benchmark evidence as if it reflected the project’s approved public positioning. The current main README intentionally publishes these project-measured results with explicit limits: Deepseek-V4-Pro, one run per case, an initial sanity signal rather than a fixed ranking, with fuller details to follow. We will keep those claims and their qualifiers.

PR #2147, opened 2026-08-07
fix: mark truncated assistant replies as incomplete instead of complete

### LINKED ISSUE #1980: [Bug]: Incomplete assistant reply is persisted as `complete` when the model output stream ends prematurely
### Summary

An assistant reply stopped mid-generation (the text ends at the title of list item 4, with the markdown list left unclosed), but the session persistence layer recorded the turn as `finishReason=complete` / `status=completed` / `hasFinalResponse=true`. The UI shows no "reply may be incomplete" indication, so the user cannot distinguish an intentional stop from a transport/generation failure. This is a silent failure with the same observability gap as session freeze / review freeze issues.

**Expected behavior**: when the model output stream ends prematurely (provider-side interruption, timeout, connection drop, or max_tokens cut), the session pipeline should record a non-`complete` finish reason (`length` / `interrupted` / `error`), persist the turn as incomplete, and let the UI surface a hint (e.g. "reply may be incomplete") instead of silently marking it complete.

### Area

Agent runtime / core (session persistence and `finishReason` determination); AI provider / model adapter (premature stream termination).

### Reproduction or evidence

Hard to reproduce deterministically (it depends on the underlying model stream ending early), but this is a real occurrence with complete local evidence from the persisted turn data (no private data included):

1. **Phenomenon**: the assistant reply was cut mid-sentence — the text stops at `4. **压缩/总结(co`, right after the "能结构性解决" table and list items 1-3 of "解决不了的(方案边界)". Item 4 has a title only, no body; the markdown list is unclosed.

2. **Persisted evidence** (local `sessions/<session-id>/turns/turn-0004.json`):
   - `textItems[0].content` is only **865 characters**, ending at `4. **压缩/总结(co`;
   - the same modelRound's `thinkingItems` are **complete**, containing the full analysis for items 8, 9, 10 and the conclusion — proving the model finished reasoning and had a full answer ready, but the text stream terminated mid-generation;
   - the turn records `finishReason=complete`, `status=completed`, `hasFinalResponse=true` — no truncation or error marker at all;
   - `tokenUsage.outputTokens=1776` (including thinking); the text is only a fraction of that, so the output stream stopped far earlier than a normal completion point.

3. **Control group**: other turns in the same session with much longer replies (e.g. 4908 and 2787 characters) completed normally. This shows it is not "long replies always get cut"; it is an intermittent premature stream termination.

### Suggested fix

1. When the response stream ends early (missing stream-end frame, `finish_reason` = `length`/`interrupted`, or visibly unclosed text), mark the round as `interrupted`/`incomplete` instead of `complete`.
2. Keep `finishReason` semantics distinct: `complete` / `length` / `interrupted` / `error`, and retain the termination reason in the round state or `tokenUsage`.
3. Surface a "reply may be incomplete" hint in the UI and allow the user to regenerate or continue.
4. Optionally add a persistence-layer integrity check (e.g. warn when thinking is complete but the text is clearly truncated).

### Environment, if relevant

- BitFun v0.2.15 (2026-07-31 release), Windows desktop app, built-in personal-assistant session.
- Not tied to a specific model/provider; any provider stream that ends early during a long output can trigger this.

xielixing:
## Fix: Mark truncated assistant replies as incomplete instead of complete

Closes #1980

### Problem

When the model output stream ends prematurely (provider-side interruption, timeout, connection drop, or `max_tokens` cut), the execution engine marks the dialog turn as `complete` / `hasFinalResponse=true` instead of `interrupted`/`incomplete`. This causes the frontend to persist and display an incomplete assistant reply as if it were a complete, final response.

### Root Cause

In [`execution_engine.rs`](src/crates/assembly/core/src/agentic/execution/execution_engine.rs), three code paths contribute to the bug:

1. **`partial_truncated` marks as final (line ~4464):** When continuation attempts are exhausted after a partial recovery, `finalization_reason` is set to `"partial_truncated"`, but the code then sets `has_final_response = true` — treating the truncated text as a complete final response.

2. **Success calculation excludes `partial_truncated` (line ~4480):** The `success` variable is `has_final_response || matches!(effective_finish_reason, "max_rounds" | "repeated_tool_failures")`. Since `"partial_truncated"` is not in the matches, and after fix #1 `has_final_response` is `false`, the turn would be marked as failed (`success=false`), which is too harsh — the turn did produce partial output.

3. **Cancellation path leaves `finalization_reason = None` (line ~4194):** When `should_continue_after_partial_response(reason)` returns `false` (reason contains "cancelled"), the code breaks immediately without setting `finalization_reason`, so it defaults to `None` → `effective_finish_reason = "complete"` → `has_final_response = true`. A cancelled stream with partial text is reported as complete.

### Fix

Three surgical changes:

1. **Line ~4465:** `has_final_response = true` → `has_final_response = false` for the `partial_truncated` branch. A truncated response is not a complete final response.

2. **Line ~4483:** Add `"partial_truncated"` to the `matches!` list for `success`. The turn produced partial output, so `success=true` with `has_final_response=false` tells the UI: "the turn produced output, but it is incomplete."

3. **Line ~4199:** Set `finalization_reason = Some("partial_truncated")` before `break` in the cancellation path. This ensures a cancelled stream with partial text is reported as `partial_truncated` (incomplete) rather than `complete`.

### Event Flow After Fix

The `DialogTurnCompleted` event now carries for truncated turns:
- `finish_reason: "partial_truncated"` (was `"complete"`)
- `has_final_response: false` (was `true`)
- `success: true` (was `true` for path 1, `false` would have been too harsh)
- `partial_recovery_reason: <actual reason>` (unchanged — e.g., "idle_timeout", "cancelled")

The frontend projection (`frontend_projection.rs`) forwards these fields as `finishReason`, `hasFinalResponse`, `success`, and `partialRecoveryReason`, allowing the UI to distinguish incomplete turns from complete ones.

### Validation

- `cargo build -p bitfun-core --lib` — compiles clean (1 pre-existing warning, unrelated)
- `cargo test -p bitfun-core --lib execution::execution_engine` — all 38 existing tests pass

### DIFF
--- a/pr-body-1980.md
+++ b/pr-body-1980.md
@@ -1,0 +1,42 @@
+## Fix: Mark truncated assistant replies as incomplete instead of complete
+
+Closes #1980
+
+### Problem
+
+When the model output stream ends prematurely (provider-side interruption, timeout, connection drop, or `max_tokens` cut), the execution engine marks the dialog turn as `complete` / `hasFinalResponse=true` instead of `interrupted`/`incomplete`. This causes the frontend to persist and display an incomplete assistant reply as if it were a complete, final response.
+
+### Root Cause
+
+In [`execution_engine.rs`](src/crates/assembly/core/src/agentic/execution/execution_engine.rs), three code paths contribute to the bug:
+
+1. **`partial_truncated` marks as final (line ~4464):** When continuation attempts are exhausted after a partial recovery, `finalization_reason` is set to `"partial_truncated"`, but the code then sets `has_final_response = true` — treating the truncated text as a complete final response.
+
+2. **Success calculation excludes `partial_truncated` (line ~4480):** The `success` variable is `has_final_response || matches!(effective_finish_reason, "max_rounds" | "repeated_tool_failures")`. Since `"partial_truncated"` is not in the matches, and after fix #1 `has_final_response` is `false`, the turn would be marked as failed (`success=false`), which is too harsh — the turn did produce partial output.
+
+3. **Cancellation path leaves `finalization_reason = None` (line ~4194):** When `should_continue_after_partial_response(reason)` returns `false` (reason contains "cancelled"), the code breaks immediately without setting `finalization_reason`, so it defaults to `None` → `effective_finish_reason = "complete"` → `has_final_response = true`. A cancelled stream with partial text is reported as complete.
+
+### Fix
+
+Three surgical changes:
+
+1. **Line ~4465:** `has_final_response = true` → `has_final_response = false` for the `partial_truncated` branch. A truncated response is not a complete final response.
+
+2. **Line ~4483:** Add `"partial_truncated"` to the `matches!` list for `success`. The turn produced partial output, so `success=true` with `has_final_response=false` tells the UI: "the turn produced output, but it is incomplete."
+
+3. **Line ~4199:** Set `finalization_reason = Some("partial_truncated")` before `break` in the cancellation path. This ensures a cancelled stream with partial text is reported as `partial_truncated` (incomplete) rather than `complete`.
+
+### Event Flow After Fix
+
+The `DialogTurnCompleted` event now carries for truncated turns:
+- `finish_reason: "partial_truncated"` (was `"complete"`)
+- `has_final_response: false` (was `true`)
+- `success: true` (was `true` for path 1, `false` would have been too harsh)
+- `partial_recovery_reason: <actual reason>` (unchanged — e.g., "idle_timeout", "cancelled")
+
+The frontend projection (`frontend_projection.rs`) forwards these fields as `finishReason`, `hasFinalResponse`, `success`, and `partialRecoveryReason`, allowing the UI to distinguish incomplete turns from complete ones.
+
+### Validation
+
+- `cargo build -p bitfun-core --lib` — compiles clean (1 pre-existing warning, unrelated)
+- `cargo test -p bitfun-core --lib execution::execution_engine` — all 38 existing tests pass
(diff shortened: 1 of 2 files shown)


limityan (changes requested):
Request changes for commit `e0f50ecb3ad562b073c0b573f2be3c22e83640ba`.

The live `DialogTurnCompleted` event is improved, but the incomplete outcome is not yet carried through the durable runtime contract.

### 1. Persist the incomplete outcome in core

**Problem:** `ExecutionEngine` emits `finish_reason=partial_truncated` and `has_final_response=false`, but the returned `ExecutionResult.finish_reason` is still hard-coded to `FinishReason::Complete`. `persist_completed_dialog_turn` / `SessionManager::complete_dialog_turn` do not receive or persist either terminal field.

**Risk:** A connected Desktop UI may later write the event fields back, but CLI, server, SDK, cron, disconnected clients, and shutdown-before-UI-save paths persist the turn without the incomplete marker. Reloading that history still cannot distinguish the truncated reply from a completed one.

**Recommendation:** Carry the effective finish reason and `has_final_response` in `ExecutionResult`, persist them atomically in core, and add a regression test that reloads the turn without any frontend write-back.

### 2. Do not notify users that a truncated background task completed successfully

**Problem:** `partial_truncated` is explicitly included in `success=true`, while the completion notification policy only treats `success === false` as a failure. The background notification therefore still says the task finished, and the in-chat notice falls back to the generic abnormal-ending copy instead of saying that the reply is incomplete.

**Risk:** The main user-facing symptom remains misleading for background work even though `has_final_response=false` is available.

**Recommendation:** Make notification and turn-notice policy consider `finishReason=partial_truncated` / `hasFinalResponse=false`, provide explicit incomplete-response copy, and cover both policies with tests.

### 3. Remove the PR-body scratch file

`pr-body-1980.md` duplicates this PR description and is an unrelated generated scratch artifact. Please remove it from the commit.

### Verification status

- `git diff --check`, repository hygiene, and changed-file `rustfmt --check` passed locally.
- The exact-head CI run is `action_required` and contains no jobs/check runs, so there is no remote passing evidence yet.
- Targeted Cargo tests did not complete locally because of shared compilation/lock contention; I am not treating the timeout as either a pass or a failure.

PR #2159, opened 2026-08-07
fix(flow-chat): drain stuck bottom whitespace after subagent card collapses

### LINKED ISSUE #2154: FlowChat: bottom whitespace grows after subagent card collapses and only drains on manual scrolling
## Problem

After a subagent (Task) card collapses, the chat viewport keeps a large block of blank space below the last message. The blank area **grows with each subagent card event** and only drains after the user manually scrubs the scroll wheel up and down repeatedly (each round trip consumes only the upward-scroll distance).

Screenshot: the failed "Explore" card is fully collapsed (compact header), the scrollbar thumb sits at the physical bottom, yet ~750-800px of empty background separates the message from the input bar.

## Root cause

Collapse-intent pre-compensation adds provisional footer space (`collapse.px`) before the card shrinks. When the user is at the physical bottom, every settlement path dead-locks:

- `settleCollapseReservationForViewport` / retained quiet settlement compute the geometric minimum for the *current* `scrollTop`, but the synthetic footer itself holds `scrollTop` at the physical bottom, so the minimum recomputes exactly the current reservation (e.g. 786px stays 786px).
- The scroll handler's per-scroll consumption never fires at the physical bottom because `scrollDelta` is 0 (scrollTop is capped by the footer).
- The idle reservation audit settles to the same value, so it cannot help either.

Multiple collapse intents (e.g. several parallel subagent failures) stack provisional estimates on the unsettled reservation (`base = current total compensation`), which is why the whitespace grows monotonically during the session.

## Fix

`convergeBottomReservationsToContentBottom`: when the scroller sits at the physical bottom (within 2px) with no `preserving-element` / `pinned-item` viewport owner and no active collapse intent, clear collapse + pin reservations in one synchronous footer update and fall `scrollTop` to the new content bottom (the delayed browser clamp compensation was hiding). Wired into:

- retained quiet settlement
- collapse-intent finalization (non-pinned strategies)
- idle reservation audit (bottom users only; mid-list readers keep their reading anchor)
- scroll handler: any scroll event at the physical bottom is bottom intent - one pass replaces per-scrub consumption
- collapse-intent accumulation guard: a new provisional estimate no longer stacks on a settled bottom reservation
- input-stack-shrink drain

Also: bound collapse-intent finalize retries (20) so a virtualized sticky-pin target cannot keep the intent active (and auto-follow suspended) forever; intent-window growth consumption is capped at the measured protection line so subagent growth no longer sediments into footer whitespace.

## Verification

Component tests cover bottom convergence (finalize / scroll / audit), negative cases (mid-list, pinned-item, preserving-element must not clear reservations), non-stacking consecutive intents, growth protection during an active intent, and bounded force-settle at mid-list. Requires manual verification with a real streaming session.

YodonTan:
Fixes #2154

## Problem

After a subagent (Task) card collapses, the chat viewport keeps a large block of blank space below the last message. The blank area grows with each subagent card event and only drains after manually scrubbing the scroll wheel (each round trip consumes only the upward-scroll distance). The scrollbar thumb sits at the physical bottom while hundreds of pixels of empty background separate the content from the input bar.

## Root cause

Collapse-intent pre-compensation adds provisional footer space (`collapse.px`) before the card shrinks. When the user is at the physical bottom, every settlement path dead-locks:

- `settleCollapseReservationForViewport` / retained quiet settlement compute the geometric minimum for the *current* `scrollTop`, but the synthetic footer itself holds `scrollTop` at the physical bottom, so the minimum recomputes exactly the current reservation (e.g. 786px stays 786px).
- The scroll handler's per-scroll consumption never fires at the physical bottom because `scrollDelta` is 0 (scrollTop is capped by the footer).
- The idle reservation audit settles to the same value, so it cannot help either.

Multiple collapse intents (e.g. several parallel subagent failures) stack provisional estimates on the unsettled reservation (`base = current total compensation`), which is why the whitespace grows monotonically during the session. Repeated up/down scrub cycles only consumed the upward-scroll distance per round trip.

## Changes

- **`convergeBottomReservationsToContentBottom`**: when the scroller sits at the physical bottom (within 2px) with no `preserving-element` / `pinned-item` viewport owner and no active collapse intent, clear collapse + pin reservations in one synchronous footer update and fall `scrollTop` to the new content bottom (the delayed browser clamp that compensation was hiding). Wired into:
  - retained quiet settlement (`settleRetainedCollapseRange`)
  - collapse-intent finalization (non-pinned strategies)
  - the idle reservation audit - bottom users only; mid-list readers keep their reading anchor (the audit no longer releases a preserved element anchor for mid-list readers)
  - the scroll handler: any scroll event at the physical bottom is bottom intent - one pass replaces per-scrub consumption
  - the collapse-intent accumulation guard: a new provisional estimate no longer stacks on a settled bottom reservation
  - the input-stack-shrink drain
- **Bound collapse-intent finalize retries** (max 20): a virtualized sticky-pin target can no longer keep the intent active (and auto-follow suspended) forever.
- **Intent-window growth consumption** is capped at the measured protection line so subagent growth no longer sediments into footer whitespace.
- **CSS**: `overflow-wrap: anywhere` + `min-width: 0` on subagent projection text and task card headers so long unbroken tokens wrap inside the card instead of widening the layout.

## Tests

New component tests in `VirtualMessageList.session-boundary.test.tsx`:
- bottom convergence on finalize / scroll / idle audit at the physical bottom
- negative cases: mid-list readers, pinned-item and preserving-element viewports must NOT clear reservations
- consecutive collapse intents do not stack provisional estimates
- growth during an active collapse intent only consumes provisional inflation (protection line kept)
- bounded force-settle to the geometric minimum at mid-list
- `data-reservation-px` footer diagnostics

All flow_chat modern + tool-cards tests pass (376), `type-check:web` and ESLint are clean.

## Manual verification

Verified with a real streaming session: long-title subagent failures collapse without leaving bottom whitespace, repeated failures no longer accumulate blank space, and the viewport settles to the content bottom without scrub-wheel workarounds.

### DIFF
--- a/src/web-ui/src/flow_chat/components/modern/FLOWCHAT_SCROLL_STABILITY.md
+++ b/src/web-ui/src/flow_chat/components/modern/FLOWCHAT_SCROLL_STABILITY.md
@@ -114,6 +114,15 @@ measurement convergence has no such owner. An idle measurement with no owner
 must only rebase the measured height and must never create Footer space from a
 negative `scrollHeight` delta alone.
 
+Owners are bounded by an idle reservation audit: while no streaming, collapse
+intent, retained anchor, or viewport owner is active and the user has been idle
+past `RETAINED_COLLAPSE_RELEASE_QUIET_MS`, `VirtualMessageList` settles the
+collapse reservation to its geometric minimum for the current scroll position
+and drops sticky pins whose target is no longer the latest turn. This is a
+safety net, not a release mechanism: a legitimate owner that still needs its
+range keeps it, but a stuck owner can no longer accumulate synthetic tail
+whitespace across audit intervals.
+
 Reservation state is ref-owned first and mirrored into React state. A Virtuoso
 Footer remount must synchronously read the ref-owned value; otherwise one stale
 React commit can remove exactly the reserved scroll range for a frame.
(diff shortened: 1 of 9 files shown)


wsp1911:
flowchat正在重构，暂不接受合入

PR #2187, opened 2026-08-10
fix(ci): generate web bindings in nightly build

wgqqqqq:
## Summary

- Generate ts-rs Web API bindings before the nightly Web UI type-check.
- Add a GitHub workflow contract test that enforces generation before type-checking.

## Verification

- pnpm run check:github-config
- node --test scripts/check-github-config.test.mjs
- git diff --check

AI-assisted change; tested with the focused GitHub workflow checks.


wgqqqqq:
Fork nightly validation: https://github.com/wgqqqqq/BitFun/actions/runs/31364998581

All 5 package platforms passed Generate web API bindings, Type-check web UI, and Patch nightly version:
- macos-x64
- macos-arm64
- windows-x64
- linux-arm64
- linux-x64

The full Tauri packaging jobs are still running. The PR Frontend Build and Rust Build Checks passed; the unrelated CLI test jobs hit the existing terminal startup prompt timeout.

PR #2213, opened 2026-08-11
fix(scene-bar): keep scene tab close button under cursor on press

### LINKED ISSUE #2210: [Bug]: 顶部标签页栏关闭按钮功能异常
### Summary

顶部标签页栏点击关闭按钮，有概率无法关闭，实测结果似乎是触发了标签页切换

### Area

Desktop app

### Reproduction or evidence

https://github.com/user-attachments/assets/dcb0e17b-2000-4fb0-b955-2b0e071c06bf

操作过程如视频所示，概率复现。

### Environment, if relevant

BitFun  Release 0.2.17 
OS：macOS Tahoe 26.5.2

linzhary:
## Summary

Fixes #2210 — the top SceneBar scene tab close button sometimes fails to close and instead activates the tab.

Root cause: the whole scene tab scales `0.985` around its center on `:active` (120ms). The close button is absolutely positioned (`right: 6px`) inside the tab, so the scale drags it left by `0.015 × (width/2 − 15)px` (≈4–8px on typical scene tab widths). The mouseup hit-test then lands outside the button, the click bubbles to the tab, and the tab is activated instead of closed. The escape is larger on wider tabs and timing-sensitive (releasing during the 120ms scale misses more often).

## Fix

- `SceneTab.tsx`: measure the tab width and expose the close button's scale drift as `--bitfun-scene-tab-close-shift`.
- `SceneBar.scss`: on `:active`, counter-translate the close button with the same `120ms cubic-bezier(0.23, 1, 0.32, 1)` easing as the tab scale, so the press animation is fully preserved while the button stays under the cursor. The button is also pinned against the global motion baseline 1px drop on `:active`. `:only-child` and `prefers-reduced-motion` paths keep the button at its base position.

## Verification

- `pnpm run type-check:web` passes.
- SCSS compiles cleanly.
- Manual verification recommended: open 2+ scene tabs in desktop dev, press the close button at its right edge and release quickly — the button no longer shifts and the first click closes the tab.

Note: this PR is AI-assisted; testing level: lightly tested.

### DIFF
--- a/src/web-ui/src/app/components/SceneBar/SceneBar.scss
+++ b/src/web-ui/src/app/components/SceneBar/SceneBar.scss
@@ -49,7 +49,7 @@
 
 .bitfun-scene-tab {
   // Measured by SceneTab and overridden per-tab; 0 until first measurement.
-  --scene-tab-close-shift: 0px;
+  --bitfun-scene-tab-close-shift: 0px;
   position: relative;
   display: flex;
   align-items: center;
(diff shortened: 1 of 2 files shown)


kev1n77:
感谢贡献，由于代码有较大更新，相关问题已在PR #2741 中修复，此PR关闭

PR #2239, opened 2026-08-12
fix(git): recover untrusted repository operations with a confirmation flow

guantw:
## Summary

Recover Git operations blocked by dubious-ownership rejection. BitFun now classifies the owner guard as a typed, recoverable error and gives the user one explicit consent flow that writes `safe.directory` and completes the blocked command across local, remote, and peer surfaces.

- Detect libgit2 / Git CLI ownership rejections at the service boundary and raise a typed error instead of an opaque failure.
- Confirm the repository root through a unified consent flow that writes `safe.directory` and replays the blocked operation once.
- Keep probes read-only and non-interactive paths prompt-free; never replay side-effecting commands without consent.
- Apply trust on the machine the user is on: remote and peer surfaces route the diagnosis and the manual command to the controller, while the write stays in the execution domain.
- Provide an authorization entry on the Git scene trust-required state, and manual-resolution guidance for uncertain or unsupported outcomes.
- Serialize Deep Review target Git reads around a single trust decision, and normalize prompt dedupe keys per platform.

## Problem

On Windows, handing a repository folder to another account (e.g. `system`) makes Git refuse to open it with "detected dubious ownership" unless the root is listed in `safe.directory`. BitFun surfaced that as an opaque command failure: the Git scene showed a static "not a repository" state with no recovery path, and read-only flows such as Review looked indistinguishable from a missing repository.

## Analysis

Git (since 2.35.2) and libgit2 refuse to operate on a repository whose directory owner differs from the current user unless its root is listed in `safe.directory`; the only sanctioned relief is that configuration entry, which BitFun never wrote. Failures came back unstructured, so callers could not tell "needs trust" from a broken repository, and concurrent read-only requests raced on the decision: one request opened the prompt while a sibling completed a probe and failed fast, aborting the whole preparation even after approval.

The fix keeps ownership trust in one platform-neutral module: rejections are classified once, trust state is reported without mutation, and `safe.directory` is written only after explicit consent, then verified. Trust is never granted implicitly as a fallback of a failed operation.

## Verification

- Manual (Windows): a repository handed to `system` reopens after the trust confirmation; the Git scene leaves the trust-required state and the previously greyed-out review entry recovers.
- Frontend git/trust suites pass (GitAPI, GitTrustPromptService, GitTrustOutcomePresenter, GitStateManager, DeepReviewService, GitScene).
- Rust: `services-integrations` git tests and `remote_workspace_policy` contract tests pass; remote and peer trust commands are policy-declared and routed.

## Follow-up

- `resolve_remote_git_target` currently returns `Option` and is called from ~30 sites; switching it to `Result` would change behavior at every call site at once, so it is tracked as a documented follow-up rather than a change inside this PR.

### DIFF
--- a/docs/features/remote-workspaces.md
+++ b/docs/features/remote-workspaces.md
@@ -91,6 +91,23 @@ The configured Docker CLI remains the security boundary. BitFun does not expose
 the Docker daemon over the network or bypass the current user's Docker
 permissions.
 
+## Git ownership trust
+
+Git refuses a repository whose directory is owned by another user until the
+path is listed in the protected `safe.directory` configuration — a common shape
+for shared SSH hosts, bind-mounted trees, and containers that run as a
+different uid. BitFun classifies that refusal as its own state instead of
+reporting "not a repository", so Git-backed surfaces can explain the wall and
+the Review launch fails with a specific reason.
+
+The decision itself belongs to the machine that owns the repository. For a
+remote workspace BitFun reports the state and hands over the exact command
+(`git config --global --add safe.directory "<path>"`) to run on the remote
+host; it never writes the remote user's global Git configuration, and it never
+grants trust implicitly as a fallback of a failed read. The same rule applies
+in Peer Device Mode: the read-only probe answers for a controller, granting is
+refused on the peer host.
+
 ## Upgrade compatibility
 
 Existing SSH profiles remain plain SSH targets because the new `proxyJump` and
--- a/src/apps/cli/src/dispatch/mod.rs
+++ b/src/apps/cli/src/dispatch/mod.rs
@@ -11,6 +11,7 @@ use std::process::Command;
 use anyhow::{anyhow, bail, Context, Result};
 use bitfun_core::infrastructure::ai::AIClientFactory;
 use bitfun_core::service::config::{AuthConfig, GlobalConfig};
+use bitfun_core::service::git::trust;
 use serde::de::DeserializeOwned;
 
 use protocol::{
@@ -758,18 +759,23 @@ fn inspect_workspace(workspace_path: &str) -> Result<DispatchWorkspaceProbe> {
     } else {
         path
     };
-    let is_git_repository = is_directory
-        && git_output(&canonical, &["rev-parse", "--is-inside-work-tree"])
-            .is_some_and(|output| output.trim() == "true");
+    let RepositoryProbe {
+        is_git_repository,
+        trust_required,
+    } = is_directory
+        .then(|| classify_repository_probe(git_probe(&canonical, REV_PARSE_INSIDE_WORK_TREE)))
+        .unwrap_or_default();
     let branch = is_git_repository
         .then(|| git_output(&canonical, &["branch", "--show-current"]))
         .flatten()
         .map(|branch| branch.trim().to_string())
         .filter(|branch| !branch.is_empty());
-    let dirty = is_git_repository.then(|| {
-        git_output(&canonical, &["status", "--porcelain"])
-            .is_some_and(|output| !output.trim().is_empty())
-    });
+    // `None`, not `Some(false)`, when the read never landed: a repository we
+    // were not allowed to inspect is not a clean one.
+    let dirty = is_git_repository
+        .then(|| git_output(&canonical, &["status", "--porcelain"]))
+        .flatten()
+        .map(|output| !output.trim().is_empty());
     let (ahead, behind) = if is_git_repository {
         git_output(
             &canonical,
@@ -786,6 +792,7 @@ fn inspect_workspace(workspace_path: &str) -> Result<DispatchWorkspaceProbe> {
         exists,
         is_directory,
         is_git_repository,
+        trust_required: trust_required.then_some(true),
         branch,
         dirty,
         ahead,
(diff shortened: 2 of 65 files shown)


wgqqqqq:
Read the full diff (62 files). Static reading only — no build, no test run.

The design is right: splitting ownership rejection out of "not a repository", reporting trust read-only, writing `safe.directory` only after explicit consent, and re-probing to verify. `RepositoryUntrusted` → `PermissionDenied` (not `NotFound`), `is_git_repository` answering `true` for a rejected repo in all three domains, and `withGitRepositoryTrustRecovery` wrapping reads only are all the right calls.

Four things I'd fix before merge.

**1. `normalize_trust_path` rewrites backslashes unconditionally — `trust.rs:117`**

```rust
let mut value = trimmed.replace('\\', "/");
```

Backslash is a legal filename character on POSIX — this repo already guards that (`targetResolver.test.ts:528`, "preserves a Unix literal backslash"). `/srv/we\ird/repo` becomes `/srv/we/ird/repo`, gets written to `safe.directory` as an entry that can never match, and the manual command names a directory that doesn't exist. Suggest gating on Windows / a drive-letter or UNC shape. Same pattern in `pathUtils.ts:148,168` (map keys only, lower impact).

**2. Path extraction takes the first quote pair — `trust.rs:147-156`**

`/srv/a'b/repo` parses as `/srv/a`. Not cosmetic: `classify_command_failure:248` gives the parsed path priority over the caller's, and it reaches `trust_repository_with_env:491` — so `git config --global --add safe.directory /srv/a` lands in the user's config, is never reclaimed, and doesn't resolve the rejection. Suggest taking the last quote on the line, or validating the result and falling back to the caller's path.

**3. No guard against `safe.directory` glob semantics — `trust.rs:486-500`**

`*` disables the ownership check entirely; a value ending in `/*` trusts everything beneath. A directory can be named `*` on POSIX, so granting a repo at `/tmp/*` silently trusts every repository under `/tmp` — well beyond what the dialog asked for. Rejecting those candidates and routing them to `manual_command` is cheap.

Related: my machine had `safe.directory = *` globally, which makes this whole feature unreachable. Might be worth surfacing that state on the trust screen.

**4. `ChatInputWorkspaceStrip` still reports "not a Git repository"**

`ChatInputWorkspaceStrip.tsx:169-176` reads `isRepository` from `useGitState(['basic'])`, but `GitStateManager.refreshBasicAndStatus` only sets it `true` after `gitAPI.getStatus()` succeeds — the exact call the ownership check blocks. The badge falls through to `flow-chat.json:577` `"非 Git 仓库或当前无分支"`, while the new `repositoryUntrusted` string (`:1191`) is wired only to `ReviewService.ts:81`. The misleading message this PR exists to remove survives on the surface closest to the user.

---

Smaller:

- `shell_quote_pretty`'s bare branch admits non-ASCII control and whitespace (U+0085, U+00A0, U+3000) via `!c.is_ascii()`, while the double-quote branch screens controls — `trust.rs:209-212` vs `:216`. Adding `&& !c.is_control() && !c.is_whitespace()` aligns them.
- `probeRepositoryTrust` (`GitStateManager.ts:592`) exists for the CLI peer, whose `git_get_status` always fails — so every refresh pays a fixed extra round-trip with no caching or backoff.
- `stableErrorPayload` (`TauriCommandError.ts:112-137`) went from 4 to 5 candidates per node but kept the budget at 12, and `inspected` counts `undefined` dequeues. Deeply wrapped web-mode errors truncate more easily now.
- `resolve_remote_git_target` `Option` → `Result` (`git_api.rs:71-89`) is correct, but it changes behavior at all 30 call sites at once — worth a line in the PR description.
- When the rejected path ends in `/.git`, `trust_candidates` writes the gitdir entry before probing whether the worktree entry would have sufficed (`trust.rs:288-301`, `486-499`); the extra entry stays in the global config.
- Remote `trusted` branch returns `stdout.trim()` raw (`git_api.rs:641-646`) while the local one normalizes — align after #1.
- The "consent precedes `git_trust_repository`" invariant lives entirely in the frontend; the backend has no consent token. Matches existing convention, but given this writes global config and lets Git run hooks from someone else's directory, worth stating explicitly.

---

Verification: I confirmed the backend classification empirically on macOS against a real `root:wheel`-owned repo — `open()` returns `RepositoryUntrusted` with the path populated, and `is_git_repository` answers `true`.

guantw:
> Read the full diff (62 files). Static reading only — no build, no test run.
> 
> The design is right: splitting ownership rejection out of "not a repository", reporting trust read-only, writing `safe.directory` only after explicit consent, and re-probing to verify. `RepositoryUntrusted` → `PermissionDenied` (not `NotFound`), `is_git_repository` answering `true` for a rejected repo in all three domains, and `withGitRepositoryTrustRecovery` wrapping reads only are all the right calls.
> 
> Four things I'd fix before merge.
> 
> **1. `normalize_trust_path` rewrites backslashes unconditionally — `trust.rs:117`**
> 
> ```rust
> let mut value = trimmed.replace('\\', "/");
> ```
> 
> Backslash is a legal filename character on POSIX — this repo already guards that (`targetResolver.test.ts:528`, "preserves a Unix literal backslash"). `/srv/we\ird/repo` becomes `/srv/we/ird/repo`, gets written to `safe.directory` as an entry that can never match, and the manual command names a directory that doesn't exist. Suggest gating on Windows / a drive-letter or UNC shape. Same pattern in `pathUtils.ts:148,168` (map keys only, lower impact).
> 
> **2. Path extraction takes the first quote pair — `trust.rs:147-156`**
> 
> `/srv/a'b/repo` parses as `/srv/a`. Not cosmetic: `classify_command_failure:248` gives the parsed path priority over the caller's, and it reaches `trust_repository_with_env:491` — so `git config --global --add safe.directory /srv/a` lands in the user's config, is never reclaimed, and doesn't resolve the rejection. Suggest taking the last quote on the line, or validating the result and falling back to the caller's path.
> 
> **3. No guard against `safe.directory` glob semantics — `trust.rs:486-500`**
> 
> `*` disables the ownership check entirely; a value ending in `/*` trusts everything beneath. A directory can be named `*` on POSIX, so granting a repo at `/tmp/*` silently trusts every repository under `/tmp` — well beyond what the dialog asked for. Rejecting those candidates and routing them to `manual_command` is cheap.
> 
> Related: my machine had `safe.directory = *` globally, which makes this whole feature unreachable. Might be worth surfacing that state on the trust screen.
> 
> **4. `ChatInputWorkspaceStrip` still reports "not a Git repository"**
> 
> `ChatInputWorkspaceStrip.tsx:169-176` reads `isRepository` from `useGitState(['basic'])`, but `GitStateManager.refreshBasicAndStatus` only sets it `true` after `gitAPI.getStatus()` succeeds — the exact call the ownership check blocks. The badge falls through to `flow-chat.json:577` `"非 Git 仓库或当前无分支"`, while the new `repositoryUntrusted` string (`:1191`) is wired only to `ReviewService.ts:81`. The misleading message this PR exists to remove survives on the surface closest to the user.
> 
> Smaller:
> 
> * `shell_quote_pretty`'s bare branch admits non-ASCII control and whitespace (U+0085, U+00A0, U+3000) via `!c.is_ascii()`, while the double-quote branch screens controls — `trust.rs:209-212` vs `:216`. Adding `&& !c.is_control() && !c.is_whitespace()` aligns them.
> * `probeRepositoryTrust` (`GitStateManager.ts:592`) exists for the CLI peer, whose `git_get_status` always fails — so every refresh pays a fixed extra round-trip with no caching or backoff.
> * `stableErrorPayload` (`TauriCommandError.ts:112-137`) went from 4 to 5 candidates per node but kept the budget at 12, and `inspected` counts `undefined` dequeues. Deeply wrapped web-mode errors truncate more easily now.
> * `resolve_remote_git_target` `Option` → `Result` (`git_api.rs:71-89`) is correct, but it changes behavior at all 30 call sites at once — worth a line in the PR description.
> * When the rejected path ends in `/.git`, `trust_candidates` writes the gitdir entry before probing whether the worktree entry would have sufficed (`trust.rs:288-301`, `486-499`); the extra entry stays in the global config.
> * Remote `trusted` branch returns `stdout.trim()` raw (`git_api.rs:641-646`) while the local one normalizes — align after [docs: add product screenshot #1](https://github.com/GCWing/BitFun/pull/1).
> * The "consent precedes `git_trust_repository`" invariant lives entirely in the frontend; the backend has no consent token. Matches existing convention, but given this writes global config and lets Git run hooks from someone else's directory, worth stating explicitly.
> 
> Verification: I confirmed the backend classification empirically on macOS against a real `root:wheel`-owned repo — `open()` returns `RepositoryUntrusted` with the path populated, and `is_git_repository` answers `true`.

All four blocking items and most smaller ones are fixed.

Declined three:
- probe cache/backoff: fires only after a failed refresh; caching defeats it on CLI peers
- `.git` ordering: the gitdir entry is the narrower one
- backend consent token: guard stays at the frontend per convention, documented at GitAPI.trustRepository

`resolve_remote_git_target` Option→Result is a documented follow-up in the PR description.

PR #2316, opened 2026-08-15
refactor(peer): give the device surface a real identity, epoch, and lifecycle

bobleer:
## Summary

Multi-device switching kept producing new failures under real use — connection drops, session list / history / streaming load failures — and #2304, #2309 and #2310 each fixed a real bug without fixing the cause. This replaces the mechanism instead of patching it again.

**The root flaw: the frontend was built for exactly one host, and multi-device was retrofitted by mutating that host underneath it.** Nothing carried a device dimension. A "switch" meant *swap a global mutable transport → wipe all product state → rebuild*.

## Type and Areas

Type: Refactor / architecture (fixes a class of bugs, not an instance)

Areas: web UI — transport, peer-device runtime, FlowChat store/manager, workspace, LSP, terminal, editor

## Motivation / Impact

Reported: with several sessions working and repeated switching between devices, "本机端和远端的会话都会出现各种问题，包括但不限于，连接中断，会话列表加载失败，会话历史加载失败，会话过程加载失败". Every one of those maps onto the same design, not onto four separate defects:

| Symptom | Cause |
|---|---|
| 会话历史加载失败 | `restoreCommandSupportKey()` labelled **every** non-SSH host `'local'`, so a capability negotiated with one peer was cached against your own machine. One older peer permanently downgraded local history restore. |
| 会话列表加载失败 | `FlowChatManager.initialize()` deduped on `[workspacePath, mode, …]`, so switching A→B→A with the same repo path **returned B's in-flight initialization to A**. Ten more caches were device-blind the same way. |
| 会话列表加载失败 / 状态错乱 | `workspaceManager.initialize()` guarded on two booleans that `clearForPeerModeSwitch()` reset — so switch N+1 cleared the guard while switch N was still awaiting, and two initializations wrote the same state concurrently. |
| 连接中断 / 永远转圈 | `PeerDeviceTransportAdapter.disconnect()` cleared its queues **without settling the promises**, so every awaiting caller hung forever. It also zeroed live concurrency counters, after which the limiter over-admitted. |
| 会话过程加载失败 | Destroy-and-rebuild with no cancellation: any in-flight work resumed against a wiped world. The three previous PRs each patched one instance of this. |

Plus a latent bug found on the way: `TerminalService.connectingPromise` was cleared *inside* the connect task — before it was ever assigned — so it latched a settled promise and every post-switch terminal reconnect would silently no-op.

## The architecture

The device surface becomes a real, versioned, isolated scope.

**1. Identity** — `deviceSurface.ts` defines `DeviceSurfaceId`, and every cache key, request-dedup key and negotiated capability is scoped by it via `surfaceScopedKey`. Workspace paths and session ids are not unique across devices; capabilities negotiated with one host say nothing about another.

**2. Per-surface state containers** — `SurfaceStateContainer` per device holds that device's sessions and history views. Switching *selects* a container; nothing is destroyed, so returning to a device renders immediately instead of reloading from scratch. Entries whose lifetime must outlive the switch that stranded them (in-flight dedup records settled in `.finally()`) stay in store-level maps keyed by surface, because a container-relative lookup there would reap the wrong device's entry.

**3. Epoch + cancellation** — each activation carries a monotonic epoch and an `AbortSignal`; stale work fails with a typed `SurfaceChangedError` that unwinds through one boundary instead of a guard at every call site. Product callers treat it as abandon-quietly, never as a load failure.

**4. Switch as an atomic state machine** — `PeerDeviceSurfaceController`, outside React. Rapid clicks coalesce to the last requested target; a superseded activation abandons rather than rolling back over the surface that replaced it. `PeerDeviceContext.tsx` drops from 579 lines to 69 — a pure binding.

**5. Connection manager outside React** — `PeerConnectionManager` owns attach/detach/keepalive with explicit `connecting | ready | degraded | lost` health, exponential-backoff reconnect, and deterministic disposal. One missed ping degrades and retries instead of tearing a working peer down.

## Verification

```
pnpm --dir src/web-ui run type-check                # clean
pnpm --dir src/web-ui run lint                      # clean
pnpm --dir src/web-ui exec vitest run               # 3448 passed (+80 new) / 14 pre-existing failures
pnpm run build:web                                  # clean (incl. appearance contract audit)
node scripts/i18n-audit.mjs                         # 0 warnings
node --test scripts/i18n-contract.test.mjs          # 37/37
theme:color-audit:all / :test / theme:visual-contract   # PASS
verify:webkit-compatibility:test                    # PASS
check:repo-hygiene / check:github-config / test:ppt-live # PASS
```

The 14 failures are `AcpAgentsConfig.test.tsx` and `modelResolution.test.ts` failing on `localStorage` being unavailable in this local Node build; they reproduce on a clean tree and pass in CI.

New tests cover the reported scenario directly, including rapid switching: coalescing A→B→C to the last target, invalidating an already-committed hydrate so the newer target proceeds, rehydrating after A→B→A, rolling a *real* failure back to the previous peer rather than local, capability isolation across surfaces, per-device metadata dedup, container preservation across switch-away-and-back, two overlapping reinitializations never running concurrently, queued transport requests settling on disposal, and switch teardown issuing no backend call.

Every new test was verified to fail with its fix reverted in place.

## Reviewer Notes

**`SurfaceChangedError` is a control-flow signal, not an error.** This is the main cross-cutting contract: anything that renders a load failure must treat `isSurfaceChangedError(e)` as abandon-quietly. `WorkspaceProvider` and the LSP initializer were updated accordingly — both would otherwise have turned an ordinary switch into a visible failure.

**Behaviour changes worth knowing:** `setTransportAdapter` retires the previous peer adapter, so a peer adapter reference held across a switch now rejects instead of talking to the wrong device — fetch it from the registry at call time. `disconnect()` is terminal. `resetForPeerModeSwitch()` now returns an empty array and no longer tears down per-session runtime.

**Invariants held:** peer history still comes only from `restore_session_view`; `loadSessionHistory` still fails closed on genuine incomplete relay imports (only the typed surface-change is exempt); catalog windowing untouched; `snapshotDropsProjectedTurnContent` gating retained; switch teardown remains frontend-only so the device you leave keeps its PTYs and language servers.

**Not verified here:** the real multi-device scenario needs the desktop app, an account, a relay and a second machine, which is not reproducible in this environment. Evidence is mechanism-level tests, not a click-through.

## Checklist

- [x] User-facing strings, docs, and locales are updated where applicable.


bobleer:
Superseded — this branch's content is already on `main` via #2314, which was opened from the same branch and merged first. `git diff upstream/main HEAD` is empty, so there is nothing to merge here. CI passing on both is independent corroboration of the same tree.

PR #2406, opened 2026-08-21
fix(config): persist review team concurrency config fields

guantw:
## Problem

The "max parallel reviewers" setting in Review settings is not persisted. After saving `max_parallel_reviewers: 1`, the config API returns success, but reopening the settings page (or restarting) shows the default `2` again, and batched Deep Review runs still launch 2 reviewers in parallel.

Root cause: `saveDefaultReviewTeamConfig` already sends `max_parallel_reviewers`, `max_retries_per_role`, `max_queue_wait_seconds`, `allow_provider_capacity_queue`, `allow_bounded_auto_retry`, and `auto_retry_elapsed_guard_seconds`, but the backend `ReviewTeamConfig` struct did not declare them. `ConfigManager::set_value_by_path` round-trips the whole `GlobalConfig` through serde, silently dropping the undeclared keys before `save_config` writes the file. Reading the config back then falls back to the struct default (`max_parallel_reviewers: 2`).

## Fix

Add the six missing persisted Review Team policy fields to `ReviewTeamConfig` with the current product defaults:

- `max_retries_per_role: 1`
- `max_parallel_reviewers: 2`
- `max_queue_wait_seconds: 1200`
- `allow_provider_capacity_queue: true`
- `allow_bounded_auto_retry: false`
- `auto_retry_elapsed_guard_seconds: 180`

The existing struct-level `#[serde(default)]` attribute keeps older persisted configs readable and supplies defaults for fields that are absent.

This also fixes a related latent bug: `max_retries_per_role` is consumed at runtime by `DeepReviewExecutionPolicy::from_config_value`, but the saved value was previously dropped on write, so a non-default value was ignored and the backend fell back to its default.

## Tests

- `review_team_policy_config_survives_service_restart` - end-to-end ConfigService write, on-disk assertion, service restart, and read-back.
- `preserves_review_team_concurrency_fields_through_config_round_trip` - verifies all six fields survive serialization and deserialization.
- `missing_review_team_concurrency_fields_use_product_defaults` - verifies older persisted config shapes remain readable and omitted fields receive product defaults.

Verified: `cargo test -p bitfun-core --no-default-features --lib -- service::config::` (83 tests, including the 3 new ones) passes.

### DIFF
--- a/src/crates/assembly/core/src/service/config/service.rs
+++ b/src/crates/assembly/core/src/service/config/service.rs
@@ -637,6 +637,90 @@ mod tests {
     }
 
     #[tokio::test]
+    async fn review_team_policy_config_survives_service_restart() {
+        let dir = tempfile::tempdir().expect("tempdir");
+        let path_manager = Arc::new(PathManager::with_user_root_for_tests(
+            dir.path().join("review-team-concurrency"),
+        ));
+        let settings = || ConfigManagerSettings {
+            path_manager: Some(path_manager.clone()),
+            auto_save: true,
+            backup_count: 0,
+        };
+
+        let service = ConfigService::with_settings(settings())
+            .await
+            .expect("config service should start");
+        service
+            .set_config(
+                "ai.review_teams.default",
+                serde_json::json!({
+                    "max_retries_per_role": 2,
+                    "max_parallel_reviewers": 1,
+                    "max_queue_wait_seconds": 45,
+                    "allow_provider_capacity_queue": false,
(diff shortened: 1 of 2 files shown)


bobleer (approved):
Reviewed the typed config round-trip, persisted defaults, legacy deserialization, and restart path. No blocking findings. Independently verified with: cargo test -p bitfun-core --no-default-features --lib -- service::config:: (83 passed). Latest CI is green.

PR #2601, opened 2026-08-28
fix(explore): backport #2565 and adapt #2595

guantw:
## Summary

- Backport #2565 so DeepReview remediation keeps the turn-admission snapshot aligned after switching the session binding to `ReviewFixer`, and admission rejects concurrent execution-affecting binding changes.
- Adapt #2595 to the `1.0.0-explore` UI instead of restoring the removed `PermissionRequestPanel`.
- Separate broad transcript routing from actionable permission ownership: delegated child requests belong to the parent surface, while direct child requests belong to the child surface.
- Reuse `ChatInputApprovalBand` inside embedded BTW/review panels so standalone child sessions can answer direct requests.
- Repair stale `@bitfun/ui` mocks and a Windows CRLF-sensitive stylesheet fixture inherited from the latest Explore base; these were exposed by full PR/local frontend suites.

## Explore-specific adaptation

`1.0.0-explore` moved approvals into the composer and removed the old floating permission panel. The primary composer therefore consumes only requests owned by the primary session, while `BtwSessionPanel` renders owner-only direct child requests with the shared compact approval band. Transcript pending state still uses the broader projection.

During review, I found that using the composer's effective target session would duplicate a direct child approval whenever the target switcher pointed at the open child panel. The composer is now bound to the primary session's ownership projection; the child panel exclusively owns the child request. Regression coverage locks this routing down.

## Verification

- `cargo test -p bitfun-core --no-default-features --features agent-runtime,git --lib dialog_turn_admission_rejects` — 4 passed
- `cargo test -p bitfun-core --no-default-features --features agent-runtime,git --lib agentic::coordination::coordinator::tests::review_fixer_turn_is_admitted_after_updating_a_deep_review_session_binding` — 1 passed
- Full Web UI Vitest suite — 569 files, 4131 passed
- `pnpm --dir src/web-ui run type-check`
- ESLint on changed production TypeScript/TSX files
- `pnpm run theme:color-audit:all`
- `pnpm run appearance:contract-audit`

## Remote scenarios

Permission responses continue through the existing per-session driver/mailbox abstraction, so no local fallback was introduced. Local Web UI and unit-level routing were exercised; remote workspace, remote control, Peer Device Mode, and Detached Dispatch were not end-to-end exercised in this change.


bobleer (approved):
Reviewed against the latest 1.0.0-explore base. The turn-admission route-key synchronization issue and narrow Core feature-profile compilation issue were fixed; focused Rust/frontend tests and the full CI matrix pass.

